PR #3, opened 2021-03-04
Add quality declaration documents

gbiggs:
This PR adds [quality declaration documents](https://www.ros.org/reps/rep-2004.html) and (where necessary) basic README files too all packages in `rmf_demos`.

The resource packages meet the requirements to be Quality Level 3. All source-code-containing packages are currently QL4.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+Any contribution that you make to this repository will
+be under the Apache 2 License, as dictated by that
+[license](http://www.apache.org/licenses/LICENSE-2.0.html):
+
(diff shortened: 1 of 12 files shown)


mxgrey (dismissed):
There's a broken link for `rmf_visualization` because we don't have quality declarations for that repo yet. I've opened an issue for it here: https://github.com/open-rmf/rmf_visualization/issues/19

I think we can go ahead and merge this with the assumption that the link will be fixed in the future when that quality declaration is done.

mxgrey (approved):
I had to force-push to fix a DCO complaint.

PR #13, opened 2021-03-25
Reduce tinyRobot vicinity to be less disruptive

mxgrey:
## Quality improvement

With the introduction of the responsive waiting behavior, it becomes harder for `tinyRobot1` to pass a `tinyRobot2` that is waiting at its charger. This PR reduces the vicinity so that `tinyRobot1` can pass without any problem.


Yadunund (approved):
Long overdue change! Thank you 🙇🏼

PR #14, opened 2021-03-25
Feature/add crowdsim example

tanyouliang95:
## What's here

This PR showcases an example of crowd simulation in `rmf_demos`'s `airport_terminal` world. Refer to [menge](http://gamma.cs.unc.edu/Menge/files/MengeTechReport.pdf) for more information about crowdsim. In this case, crowd_sim comes in handy to create intentional delay to any moving robot when humans walk pass its traffic lane.

Screenshot from 2021-03-25 16-36-35

This scenario is located at the centre of the `airport_terminal` map, in which the humans will interact with the moving cleaning robots according to the defined `human_lanes`

Minor side fixes:
- Support headless option for ignition, by `headless:=1`
- `rmf_panel_panel` will create a `dist` folder if the folder is not available during webpack download.

## Run Scenario

```bash
ros2 launch rmf_demos airport_terminal.launch.xml use_crowdsim:=1 headless:=0 use_ignition:=0
```
** Feel free to toggle the launch args: `use_crowdsim`, `headless`, `use_ignition`

To dispatch the cleaning robots: use dashboard (localhost:5000), send the cleaning robot to `zone_1` and `zone_3` respectively

## Notes and TODO
- Documentation on configuring `CrowdSim` with `traffic editor`: 
  - [This](https://github.com/FloodShao/crowd_simulation/blob/master/crowd_simulation_doc/crowd_simulation_usage.md) detailed doc was created by @ FloodShao
  - At some point, we will need to update and consolidate this doc to somewhere (maybe `multirobotbook`).

- Further simplify crowd sim configuration
  - Certain part of the configuration can be further simplified (e.g. `agent_groups` x, y coordinations, let `goal_set` be part of `states`...)
  - TODO: Cherry-pick features from https://github.com/osrf/traffic_editor/pull/279

-  Issue: when multiple agents in `agent_groups` spawn at the same x y coordinate. 
    - The noticed behaviour is when more than 10 external_agents are spawned, `menge` will fail silently during initialization. This is due to agents being spawned on the same position. (brief analysis)
    - Temp solution: 2 `agents_name` are created to accommodate more than 10 external agents. Refer [here](https://github.com/open-rmf/rmf_demos/blob/62cec8fc2c511b7651e63f2e5fba396a673b4b88/rmf_demos_maps/maps/airport_terminal/airport_terminal.building.yaml#L3-L10)


luca-della-vedova:
I am a bit worried about starting to add binaries to the repo. The [MaleVisitorPhone](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/MaleVisitorPhone) model is on fuel, is there any way to just download it instead?
I guess a hacky quick fix could be a script on compilation, a more proper one would be to integrate with pit_crew.

tanyouliang95:
Yeap, I was also quite hesitant to add the `MaleVisitorPhone` locally. Still, treating the actor as a non-static object which is similar to other robots justify the reason for having the actor in `rmf_demos_assets`. 

Of course, if `pitcrew` is prefered, a (should be) quick fix is to add a few lines of code [here](https://github.com/osrf/traffic_editor/blob/b1ee3b7596c6c73766dd6440f76bd2defb63c5a6/building_map_tools/building_map_model_downloader/building_map_model_downloader.py#L59) on `building_map_tools`.

luca-della-vedova:
The main difference with robot models is that those are not uploaded in fuel so there is no risk of them getting out of sync.
On the other hand I suspect that integrating with pit_crew wouldn't be super simple but I'm not too familiar with it.
My worry is the recurring issue of having two different version of the same models (fuel and this repo). There was a bit of a hacky solution implemented before [here](https://github.com/osrf/rmf_demos/pull/174/commits/1adc1a3fb30b2eb6ef9c49ac2bdd50a4a2027b3d), do you think it would be a reasonable hack or better to just include it in the repo?

luca-della-vedova:
Crowdsim is failing to find a solution at startup "once in a while", and all humans stop and start looking at their phone:

image

I suspect it might be an issue of too many humans / some goal sets not having enough capacity but I'm not 100% sure.
Did you observe this behavior?

tanyouliang95:
@luca-della-vedova This behaviour happened occasionally, might due to what you suspected: the capacity of each `goal_sets`. After https://github.com/open-rmf/rmf_demos/pull/14/commits/2565a6c60de51719cec21a1d4dcd1f32cdbf138a, increasing the capacity seems to solve the problem (after running some while on my machine). Let me know if this happens again on your test system.

rmf_demos_panel/rmf_demos_panel/api_server.py
```diff
@@ -52,7 +52,7 @@
 
 class DispatcherClient(Node):
     def __init__(self):
-        super().__init__('dispatcher_client')
+        super().__init__('api_server')
```
luca-della-vedova, line 55: Hem what are these changes about?

rmf_demos_panel/rmf_demos_panel/gui_server.py
```diff
@@ -37,13 +37,17 @@ def download_webpack():
...
     if os.path.isfile(bundle_path):
         print("BundleJS File exists, SKIP!")
     else:
         print(f"BundleJS File does not exist, Download it from {source_url}")
         response = requests.get(source_url)
         if response.status_code == 200:
+            dir_path = os.path.dirname(bundle_path)
+            if not os.path.isdir(dir_path):
+                print(f"Dir path: {dir_path} doesnt exist, create dir")
+                os.mkdir(dir_path)
```
luca-della-vedova, line 50: Same as above

tanyouliang95: This is related to `rmf_panel_panel`, in which to create a `dist` dir if the folder is not available. This is mentioned in the description as a minor not crucial diff. I can open another pr if this looks confusing to have it here.

luca-della-vedova: That's probably a good idea? It might be confusing in the version history to see this file change with a commit message related to crowd simulation

tanyouliang95: Removed, (same as above): https://github.com/open-rmf/rmf_demos/pull/14/commits/89270932e657136cbf8c52f2cd135b0ea5560a85

rmf_demos_panel/rmf_demos_panel/gui_server.py
```diff
@@ -37,7 +37,6 @@ def download_webpack():
     # it is advice to update the version (in '.github/workflows/gh_page.yml')
     # whenever a new feature is introduced
     source_url = "https://open-rmf.github.io/rmf_demos/v1.0.0/app.bundle.js"
-
```
luca-della-vedova, line 40: Nit, you can just do git revert to revert changes to make sure the file doesn't appear edited when we merge, even with a single line it will pollute history. Ideally the whole PR would only have 4 files changed

rmf_demos/package.xml
```diff
@@ -24,7 +24,7 @@
   <exec_depend>rmf_demos_assets</exec_depend>
   <exec_depend>rmf_robot_sim_gazebo_plugins</exec_depend>
   <exec_depend>rmf_demos_tasks</exec_depend>
-  <exec_depend>rmf_rviz_plugin</exec_depend>
+  <exec_depend>rviz2_plugin</exec_depend>
```
luca-della-vedova, line 27: Can we put this in another PR? Seems to be unrelated to crowdsim

luca-della-vedova (approved):

PR #17, opened 2021-03-29
rename dependency following recent refactor

cnboonhan:
### Fixed bug

Rename dependency to follow refactoring: https://github.com/open-rmf/rmf_traffic_editor/pull/308

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,6 +50,6 @@ jobs:
         check-latest: true
     - name: build-panel
       run: |
-        cd rmf_demos_ws/src/rmf/rmf_demos/rmf_demos_panel/rmf_demos_panel/static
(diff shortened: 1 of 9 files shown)


tanyouliang95 (approved):
Thanks!

tanyouliang95 (changes requested):
One thing that i missed out. This CI build will fail because of the removal of 
```yaml
    - name: checkout
      uses: actions/checkout@v2
      with:
        path: rmf_demos_ws/src/rmf/rmf_demos
```
from `build.yaml`, in https://github.com/open-rmf/rmf_demos/pull/15/

This is important so that we are using the current changes in this repo, instead of the `rmf.repos`.

tanyouliang95 (dismissed):

tanyouliang95 (approved):
:raised_hands:

tanyouliang95 (approved):

PR #20, opened 2021-03-30
Remove outdated traffic editor .project files

luca-della-vedova:
### Fixed bug

`.project` files for traffic_editor maps are not used anymore since the refactoring took place in [traffic_editor's #301](https://github.com/open-rmf/rmf_traffic_editor/pull/301). When opening them with traffic_editor a warning is printed and the project failes to load, `.building` files should be used instead.

### Fix applied

This PR removes the deprecated files.

### DIFF
--- a/rmf_demos_maps/maps/airport_terminal/airport_terminal.project.yaml
+++ b/rmf_demos_maps/maps/airport_terminal/airport_terminal.project.yaml
@@ -1,6 +1,0 @@
-building:
-  filename: airport_terminal.building.yaml
-name: airport_terminal
-traffic_maps:
-  {}
-version: 1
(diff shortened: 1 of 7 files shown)


luca-della-vedova:
It's also true however that the current rmf_demo_maps would actually fail building with an outdated binary release of rmf since a lot of packages just went through renaming. For example the package in the release to build maps is called `building_map_tools` while now it has been rename to `rmf_building_map_tools` https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_maps/CMakeLists.txt#L33 so compilation would actually fail because the package is not found.
Because of all the renaming that happened builds should probably be either full binary or full source until a new release is triggered.

Issue #12, opened 2021-03-24
README: inline videos are missing


codebot:
The previous repo `osrf/rmf_demos` had inline videos in the README for drive-by "is this cool" type of bling. After the migration, those are no longer here (probably to make the repo smaller?). "it would be great" to host those videos somewhere else (Google Drive?) so they show up when GitHub renders the README.

Yadunund:
@tanyouliang95 had suggested having a github repo just to store RMF related media (pics, gifs, etc). This repo will not be cloned into any workspace. Rather, the `README` in `rmf_demos` would point to the content in the media repo. But open to the idea of using GDrive too.

codebot:
I think one challenge with a github repo for media would be that it would grow large over time as we uploading new versions of the videos, unless we scrape the history and force-push it periodically. By that time I think it's basically become functionally equivalent to GDrive :man_shrugging: but perhaps with not as awesome a CDN.

tanyouliang95:
If we are still hosting the media files on github, we can also store them in a separate `media` branch, which will not consist of any past commit history. Refer to [here](https://gist.github.com/joncardasis/e6494afd538a400722545163eb2e1fa5). But still, I have no preference on where to store it.

On top of that, I also tried to update [rmf_demos readme](https://github.com/open-rmf/rmf_demos/tree/fix/readme-media-files) with this gdrive [link](https://drive.google.com/drive/folders/1xPV___Rm85o45FS2_Y04qQUvcF9x3ZLq?usp=sharing). One downside is that static gdrive img links are not verbose.

PR #21, opened 2021-04-01
Fix readme media

tanyouliang95:
Re-add the missing media files and docs in `README.md`. https://github.com/open-rmf/rmf_demos/issues/12

In order to keep the trim down the usage of media files, without using up space in the git history, we will strore all media files in a `media` [branch](https://github.com/open-rmf/rmf_demos/tree/media). This media branch will be treated as a static file hosting site. Thus whenver new media files are uploaded, we will forcefully clean all history in `media` branch.

Extra Note: Tested the gdrive method https://github.com/open-rmf/rmf_demos/commit/5a4c46016d5943e4b8433bc9c4c0d79eadabb659, wasnt ideal as the cdn links tend to be unstable (weirdly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The Robotics Middleware Framework (RMF) enables interoperability among heterogen
 
 This repository contains demonstrations of the above mentioned capabilities of RMF. It serves as a starting point for working and integrating with RMF.
 
-[<img src="http://drive.google.com/uc?export=view&id=16Qow_k0ozzcfpSij73tTffnSFtNduLvR" alt="thumbnail.png">](https://vimeo.com/405803151)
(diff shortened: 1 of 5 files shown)


codebot (commented):
Nice! But I think the links in the README are not pointing to the `media` branch, unless I'm mistaken (?) it looks like they're just trying to do a relative link to a `media` directory (?)

codebot (approved):
I don't know how those links work, but they work! So LGTM!

tanyouliang95:
It seems that you can change the branch by going one level up the dir and then append the branch name. Attached the info link in case someone will want to update the file in the future:
https://gist.github.com/joncardasis/e6494afd538a400722545163eb2e1fa5

PR #22, opened 2021-04-05
Fix out of range values for sdf compliance

luca-della-vedova:
### Fixed bug

The values for TeleportDispenser and TeleportIngestor `ambient` and `diffuse` are outside of the [0.0 - 1.0] range which is [against the specification](http://sdformat.org/spec?ver=1.7&elem=material#material_ambient) and causes runtime errors since sdformat11.

### Fix applied

Changed the colors to a neutral grey in the range of [0.0 - 1.0]


tanyouliang95 (approved):
LGTM!

tanyouliang95 (approved):

PR #25, opened 2021-04-06
Minor refactoring of api-server

tanyouliang95:
This is merely a refactoring of the `api-server` in `rmf_demos_panel`, in which some feedbacks are referred to https://github.com/open-rmf/rmf-web/pull/291. Splitting out a `dispatcher_client` helps for a future webserver switch other than using wsgi based `flask`. Will keep enhancing the current `api-server` until `rmf-web` is fully ready. 

Here all interfaces and Apis remain the same. Thus, the only test is to make sure the demos run perfectly with the gui.


Yadunund (approved):
Tested and works!

PR #30, opened 2021-04-12
clean flask-cors rosdep

tanyouliang95:
Remove `flask-cors` since not avail during `rosdep` installation

### DIFF
--- a/rmf_demos_panel/package.xml
+++ b/rmf_demos_panel/package.xml
@@ -8,8 +8,6 @@
   <license>Apache License 2.0</license>
 
   <exec_depend>python3-flask</exec_depend>
-  <exec_depend>python3-flask-cors</exec_depend>
-
   <exec_depend>rmf_fleet_msgs</exec_depend>
   <exec_depend>rmf_task_msgs</exec_depend>
   <exec_depend>rmf_demos_dashboard_resources</exec_depend>


Yadunund:
Looks like its a python2 package? It's missing definitions for `python3-flask-cors`

tanyouliang95:
Good Idea. In this case, we should add it to `rosdep`. Will keep this track open, and I will open a pr there.

Yadunund:
@tanyouliang95 there is an issue ticket open for other dependencies too. Perhaps you could address all of them?
https://github.com/open-rmf/rmf/issues/27

tanyouliang95:
Open PR: https://github.com/ros/rosdistro/pull/29086

tanyouliang95:
Will close this after added `python3-flask-cors`  to https://github.com/ros/rosdistro/blob/df4656b9b9e4485924f7b4a5deddcb12ae25b890/rosdep/python.yaml#L6020

Yadunund:
Awesome, thanks for getting these into `rosdep` @tanyouliang95!
Are the other packages `python3-shapely python3-yaml python3-requests` already available upstream?

tanyouliang95:
Yes, i think they got updated along the way. Can check it out [here](https://github.com/ros/rosdistro/blob/df4656b9b9e4485924f7b4a5deddcb12ae25b890/rosdep/python.yaml)

Issue #6, opened 2021-03-11
Use numeric input instead of dropdown for gui priority selection


tanyouliang95:
The current way of having a binary priority input (0 low and 1: high) will work for now. For future support of a "non-binary" priority input, it will be ideal to use [this](https://www.npmjs.com/package/react-numeric-input) kind of input format.

PR #33, opened 2021-04-16
fix/changes to ui panel 

vallq:
Signed-off-by: vallq <val@openrobotics.org>

### Fixed bug
#6 

### Fix applied

Updated the panel to use numeric input for the priority selection.

Other bugs / UI changes:
- Snackbar notifications did not disappear after a set interval: Update the panel to use Material UI's Snackbar component instead of Antd's message component.
- Cleaning form did not allow resubmission of the same form: Added in an attribute to the Autofill input to render an empty field with placeholder values after every submission. Essentially clearing the form to a clean state.
- Previous API calls did not yield the expected return messages: Refactored the `services.tsx` to share the same async


rmf_demos_panel/rmf_demos_panel/static/src/components/services.tsx
```diff
@@ -1,5 +1,3 @@
-import { showErrorMessage, showSuccessMessage } from './fixed-components/messages';
-
 const API_SERVER_ADD = "http://" + location.hostname + ":8080"
 const GUI_SERVER_ADD = "http://" + location.hostname + ":5000"
```
tanyouliang95, line 2: Would like to sneak in this change:
```
const GUI_SERVER_ADD = "http://" + location.host
```

tanyouliang95 (changes requested):
Thanks for making all these fixes and refactoring. Some additional feedback: 
Accidental commit of `web_server.log` file, this needs to be removed. Also, the previously mentioned issue of using numerical input for priority seems not yet implemented? Lastly, can we color code the Canceled Task progress bar to a different color? If it is not possible, maybe we can use red? (similar to failed task).

rmf_demos_panel/rmf_demos_panel/static/src/components/services.tsx
vallq: oh okay!

and about the numerical input, sorry I think I missed it out! we are looking at only 0 and 1 or a range from 0 to 1? (0, 0.1, 0.2 ... 1.0)

tanyouliang95: We can use [this](https://www.npmjs.com/package/react-numeric-input). The numerical input should be in integer, from 0-N.

vallq:
Sorry about the force pushes @tanyouliang95 , I keep missing out on the `-S` flag for commits. But I've kept the progress bar to default to a grey bar when it has been cancelled, along with disabling the cancel task button (although I expect we will change this to a revive button).

rmf_demos_panel/rmf_demos_panel/static/src/components/tasks/task-card.tsx
```diff
@@ -24,82 +26,123 @@ interface TaskCardProps {
...
+                  </Typography>
+                </Grid>
+                <Grid item xs={3}><Typography>{taskState.start_time}</Typography></Grid>
+                <Grid item xs={3}>
+                  <Typography variant="subtitle2" align="left" color="textSecondary" gutterBottom>
+                    End:
+                  </Typography>
+                </Grid>
+                <Grid item xs={3}><Typography>{taskState.end_time}</Typography></Grid>
+                <Grid item xs={12}><Button variant="outlined" onClick={() => handleCancel(taskState.task_id)} disabled={ currentTaskStatus === 'Cancelled' ? true : false }>Cancel Task</Button></Grid>
```
tanyouliang95, line 142: It will be nice if we can blur out the cancel button for `Failed` and `Completed` task card. Not sure if the fix should be here.

tanyouliang95 (changes requested):
Nice! Tested and overall i think the fixes and features are ready. Thanks alot. Just one more suggestion.

tanyouliang95 (approved):
Awesome, tested and works!

PR #35, opened 2021-04-20
Fix ignition plugin to be version independent

luca-della-vedova:
### Fixed bug

The Caddy plugin needs the ignition diff-drive system to be operated through teleop, however it was using the versioned one which means it would have to be updated at every library update (it was set for dome, and wasn't working anymore in edifice).
  
### Fix applied

The `model.sdf` file now references the unversioned file. If both a binary and a source build of Ignition Gazebo are installed and the source build should have priority, however if multiple binary versions were to be installed I'm not sure what would happen (this is not a supported case anyway).


tanyouliang95 (approved):
Tested. Caddy works on ignition. (btw the error printout for ign and gz plugin mode confused me for a while) :sweat_smile:

PR #36, opened 2021-04-20
add flag to automatically set gazebo update rate

cnboonhan:
Signed-off-by: Boon Han <charayaphan.nakorn.boon.han@gmail.com>

### Implemented feature
Proposing to add a somewhat "dirty" option to automatically set gazebo "real time update rate" to a given float. In current implementation, if the following flag is set ( for example ):
```
ros2 launch rmf_demos office.launch.xml real_time_update_rate:=0
```
then after the launch file is run, a 10 second countdown is run before the command `gz physics -u 0` is run which sets the real real update rate to 0 ( which brings the simulation to the max speed ).

The default is set to 1.0 ( real time ):
```
ros2 launch rmf_demos office.launch.xml real_time_update_rate:=1.0
```

Only implemented for classic gazebo for now as I'm unfamiliar with ignition.

### DIFF
--- a/rmf_demos/launch/simulation.launch.xml
+++ b/rmf_demos/launch/simulation.launch.xml
@@ -5,6 +5,7 @@
   <arg name="map_name" description="Name of the rmf_demos map to simulate" />
   <arg name="use_ignition" default="0" />
   <arg name="use_crowdsim" default='0'/>
+  <arg name="real_time_update_rate" default="1.0"/>
 
   <!-- Gazebo classic was chosen-->
   <group unless="$(var use_ignition)">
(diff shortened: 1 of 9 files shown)


rmf_demos/launch/simulation.launch.xml
```diff
@@ -25,6 +26,9 @@
       <env name="GAZEBO_MODEL_DATABASE_URI" value="" />
       <env name="MENGE_RESOURCE_PATH" value="$(var menge_resource_path)"/>
     </executable>
+    <executable cmd="bash -c &quot;sleep 10; gz physics -u $GAZEBO_REAL_TIME_UPDATE_RATE&quot;" output="both">
```
mxgrey, line 29: Could you elaborate on why there's a `sleep 10;` here?

cnboonhan: if I run this command without sleep, my gazebo instance seems to hang. I'm not entirely sure why, but adding a sleep seems to handle this. This is the "hacky" part of this solution. Perhaps there is a better way of dealing with this, I'd love to pick your brain on this.

PR #39, opened 2021-05-03
fix double execution of building-map custom command

tanyouliang95:
### Fixed bug

This bug is noticed from [here](https://github.com/open-rmf/rmf_traffic_editor/pull/345). The issue here is that we are running the `building_map_generator` twice for every map generation during `colcon build`, which is bad. 
Essentially, the cmake `add_custom_command()` for the `building_map_generator` is executed twice. This is due to the "double trigger" from `add_custom_target`.

### Fix applied

Remove the unnecessary `add_custom_target` line. This applies also to the build ign map process.

As the result, this will fix the mentioned issue in the attached pr, of having duplicated fields in `model.material`.

### DIFF
--- a/rmf_demos_maps/CMakeLists.txt
+++ b/rmf_demos_maps/CMakeLists.txt
@@ -26,7 +26,9 @@ foreach(path ${traffic_editor_paths})
   set(output_model_dir ${output_dir}/models)
 
   ##############################################################################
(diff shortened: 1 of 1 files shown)


luca-della-vedova (approved):
Works for me and improves map building performance by about 25% on the linked PR!

Issue #46, opened 2021-05-18
Caddy model fails when increase step size to 0.02 in gz


youliangtan:
### Description of the bug
Caddy fails when increases the step size to `0.02` in gazebo. Notice that the model will spawn back to the origin

https://user-images.githubusercontent.com/32189404/118593540-75ada580-b7da-11eb-8daf-1d024f8b149f.mp4

(not sure why the video is not viewable on firefox :disappointed: , try chrome)

### Steps to reproduce the bug

1. `ros2 launch rmf_demos airport_terminal.launch.xml`
2. Run this when launch is completed: `gz physics -s 0.02`

Observe the caddy robot. 

Occasionally will get this error when repeat the above steps:
```bash
[gzclient-14] gzclient: /build/ogre-1.9-kiU5_5/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreAxisAlignedBox.h:251: void Ogre::AxisAlignedBox::setExtents(const Ogre::Vector3&, const Ogre::Vector3&): Assertion `(min.x <= max.x && min.y <= max.y && min.z <= max.z) && "The minimum corner of the box must be less than or equal to maximum corner"' failed.
[ERROR] [gzclient-14]: process has died [pid 5
```

Note that this will not happen when using stepsize 0.01 in gazebo. On ignition, it works perfectly fine.

I tried to remove the `readonly` plugin in the `rmf_demos_assets/models/Caddy`'s `model.sdf` and spawn `caddy` at other location. But this issue still happens. Will require a further look into the model to debug on this.

mxgrey:
This is most likely an internal Gazebo/physics issue rather than anything we're doing with RMF. If I understand correctly, when model states start to get populated with NaNs, Gazebo might fall back to just placing the model at the origin.

NaNs happen when the physics can't manage to solve the constraints in a numerically stable way. Doubling the step size can easily make the physics constraints numerically unstable. I think we'll have to accept that we can't use the caddy model when we want the step size to be `0.02`. Personally I never fiddle with the step size because of precisely this kind of issue.

PR #47, opened 2021-05-20
fix caddy model and try teleop

tanyouliang95:
### Fix caddy model

Related to https://github.com/open-rmf/rmf_demos/issues/46. Seems that one of the of vehicle switch (for aesthetic) joint damping config is causing this issue. Similar to these `PolarisRanger` models in [fuel](https://app.ignitionrobotics.org/search;q=Polaris%20Ranger).

Quick fix by fixing the caddy's sdf.

For testing, 
1) launch `airport world`, 
2) use `gz physics -s 0.02`, 
3) run `ros2 run teleop_twist_keyboard teleop_twist_keyboard` to make sure caddy works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,11 @@ Non-autonomous vehicles can also be integrated with RMF provided their positions
 
 In the airport terminal map, a `Caddy` is spawned in the far right corner and can be controlled with `geometry_msgs/Twist` messages published over the `cmd_vel` topic. 
 
+Run `teleop_twist_keyboard` to control the `caddy` with your keyboard:
+```bash
+ros2 run teleop_twist_keyboard teleop_twist_keyboard
+```
+
 ![](../media/caddy.gif)
 
 ---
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -120,6 +120,11 @@ Non-autonomous vehicles can also be integrated with RMF provided their positions
 
 In the airport terminal map, a `Caddy` is spawned in the far right corner and can be controlled with `geometry_msgs/Twist` messages published over the `cmd_vel` topic. 
 
+Run `teleop_twist_keyboard` to control the `caddy` with your keyboard:
+```bash
+ros2 run teleop_twist_keyboard teleop_twist_keyboard
+```
```
luca-della-vedova, line 126: Nit, it seems we have lauch files for this, can you change the instructions to use those instead? (So it would also work for ignition if we used the `_ign` launch file)

tanyouliang95: Updated the docs. Note that the previous `airport_terminal_caddy.launch.xml`  is using a physical joystick as controller input, thus it is not included here. (since users mostly use keyboard)

luca-della-vedova (approved):
Works well, thanks for the fix!

PR #48, opened 2021-05-27
rmf_demos_dashboard_resources: Added a header logo to dashboard resources

matiasbavera:
### Implemented feature

Updated build_configurations script and added a header logo for the Dashboard

@vallq now went you pull this folder you will get the icon logo as well, and that will be accessible from the dashboard. If you want to change anything, feel free to do it

I cannot assign a reviewer, so I'm tagging @Yadunund  or @tanyouliang95  

Also, if you're ok with the content, please merge it, I don't have the permission


tanyouliang95:
Thanks for adding this. I notice that some of the config fields are no longer used by `rmf-web`. Specifically, those drop-down place selections. (Now the user will also need to manually input the `dispenser` and `ingestor` for `delivery` task on the ui). As such, those dropdown place selections are now run-time dependent on `/rmf_building_maps_msgs`. This can be quite helpful, but then there's no way for us to filter them.

As observed (correct me if I am wrong), currently those static images (e.g. logos...) are the only thing in use. For further thoughts, is there any better/preferred way to handle all these configurations between `rmf_demos` and `rmf-web`? 

p/s: there should be no pending issue with this pr.

matiasbavera:
> Thanks for adding this. I notice that some of the config fields are no longer used by `rmf-web`. Specifically, those drop-down place selections. (Now the user will also need to manually input the `dispenser` and `ingestor` for `delivery` task on the ui). As such, those dropdown place selections are now run-time dependent on `/rmf_building_maps_msgs`. This can be quite helpful, but then there's no way for us to filter them.

Yeah, It doesn't seem to be used anymore but let's ask @koonpeng  just in case.

> As observed (correct me if I am wrong), currently those static images (e.g. logos...) are the only thing in use. For further thoughts, is there any better/preferred way to handle all these configurations between `rmf_demos` and `rmf-web`?

Yeah, the logos, robots and dispensers, and more in the future. I think it depends on how many times do you want to change those logos, and do we want the end-user to be able to switch images? Or do we want to have a "friendlier" interface for the one who is deploying and change those images on runtime? I think we need to answer those questions if we want to switch approaches.

> p/s: there should be no pending issue with this pr.

TY

Yadunund (approved):
Thanks for these changes @matiasbavera. Works great.

Yadunund:
@matiasbavera looks like your commit was not GPG signed. Could you amend your commit?

koonpeng:
> Yeah, It doesn't seem to be used anymore but let's ask @koonpeng just in case.

The UI now uses the api-server to get the list of dispensers and ingestors, previously because there is no server, it couldn't know what dispensers/ingestors are available without first receiving a state update, so there is a problem where it will never know an offline dispenser/ingestor. This is workaround somewhat by the api-server as it maintains a list of dispenser/ingestors in a database, without the server, the UI will "forget" about all dispensers/ingestors everytime you refresh the page.

I think moving forward, we need a better plan on how to populate and configure these parameters, using build time config for things like dispensers, ingestors does not seem like a sensible/flexible approach.

matiasbavera:
Hey, @Yadunund can you re-trigger the workflow? It failed because of the gpg expired keys =)

image

matiasbavera:
> @matiasbavera looks like your commit was not GPG signed. Could you amend your commit?

Done! Thanks for the feedback

tanyouliang95:
Hi there, this gpg key issue is due to [this](https://discourse.ros.org/t/ros-gpg-key-expiration-incident/20669/). Should be fixed by tues US time. Meanwhile, I will try to run the setup with a different icon. Thanks

Yadunund (approved):

PR #51, opened 2021-05-31
Support non-symmetrical double swing doors

aaronchongth:
### Implemented feature

Changes to different maps with doors, after support for non-symmetrical doors, https://github.com/open-rmf/rmf_traffic_editor/pull/352.

Some changes in the PR was automatically generated after saving in `traffic-editor`.


tanyouliang95 (approved):
Tested it and also tried changing the `right_left_ratio`, the new feature works and not breaking any of the demo worlds. :+1:  Should we also add a couple of example doors with non-symmetrical profile? (Optional thou)  Thanks

PR #68, opened 2021-08-02
making models download location packaging friendly

marcoag:
Signed-off-by: Marco A. Gutierrez <marco@openrobotics.org>

### Fixed bug

Models were being downloaded to the user's home directory at (almost) compilation time, which is fine when building from sources but not so fine when trying to package the sources.
### Fix applied

Models are now downloaded to a location relative to the package so they will be included in the future packaged `.deb` file.

### DIFF
--- a/rmf_demos_maps/CMakeLists.txt
+++ b/rmf_demos_maps/CMakeLists.txt
@@ -38,12 +38,21 @@ foreach(path ${traffic_editor_paths})
     )
   else()
     message("DOWNLOADING MODELS WITH COMMAND: ros2 run rmf_building_map_tools building_map_model_downloader ${map_path}")
-    add_custom_command(
-      DEPENDS ${map_path}
-      COMMAND ros2 run rmf_building_map_tools building_map_generator gazebo ${map_path} ${output_world_path} ${output_model_dir}
-      COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ${output_model_dir}
-      OUTPUT ${output_world_path}
+    if(DEFINED ENV{RMF_DOWNLOAD_MODELS_HOME})
+      add_custom_command(
+        DEPENDS ${map_path}
+        COMMAND ros2 run rmf_building_map_tools building_map_generator gazebo ${map_path} ${output_world_path} ${output_model_dir}
+        COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ~/.gazebo/models
+        OUTPUT ${output_world_path}
       )
+    else()
+      add_custom_command(
+        DEPENDS ${map_path}
+        COMMAND ros2 run rmf_building_map_tools building_map_generator gazebo ${map_path} ${output_world_path} ${output_model_dir}
+        COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ${output_model_dir}
+        OUTPUT ${output_world_path}
+      )
+    endif()
   endif()
 
   ##############################################################################
--- a/rmf_demos_maps/CMakeLists.txt
+++ b/rmf_demos_maps/CMakeLists.txt
@@ -41,7 +41,7 @@ foreach(path ${traffic_editor_paths})
     add_custom_command(
       DEPENDS ${map_path}
       COMMAND ros2 run rmf_building_map_tools building_map_generator gazebo ${map_path} ${output_world_path} ${output_model_dir}
-      COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ~/.gazebo/models
+      COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ${output_model_dir}
       OUTPUT ${output_world_path}
       )
   endif()


rmf_demos_maps/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ foreach(path ${traffic_editor_paths})
     add_custom_command(
       DEPENDS ${map_path}
       COMMAND ros2 run rmf_building_map_tools building_map_generator gazebo ${map_path} ${output_world_path} ${output_model_dir}
-      COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ~/.gazebo/models
+      COMMAND ros2 run rmf_building_map_tools building_map_model_downloader ${map_path} -f -e ${output_model_dir}
```
aaronchongth, line 44: This will download models into the package's `install` directory, and will have to be re-downloaded if the users ever want to recompile the package after `rm -rf built/PACKAGE install/PACKAGE`.

I would suggest checking an environment variable, whether this is for source build or binary build, and changing this value accordingly. 

Pardon the pseudo-code, in this case, we would need to do something like `export BINARY_BUILD=1` or something, then in this file,
```
set(download_path '~/.gazebo/models')
if (${BINARY_BUILD})
  set(download_path ${output_model_dir})
endif()
```

aaronchongth (changes requested):
Thanks for pointing this out! Never thought this could be a potential problem. Considering that the assets will be downloaded into the binary, I would be curious how large the binary would be in the end. Is that something that we should be concerned about?

If so, we might need to rethink how we manage assets overall in demo packages.

rmf_demos_maps/CMakeLists.txt
marcoag: I'm ok with this if we have BINARY_BUILD=1 by default (maybe `PACKAGE_BUILD` is a better name?) the problem with having a non default-set environment variable is that it has to be set before packaging and to do that I have to add a patch on the release repo. I would like to avoid those are much as possible because they are exceptions that have to be treated on a case by case basis and updated when required which make them a good possible point of failure for current and future builds.

marcoag:
Yup size of the `deb` package is a concern. However I believe it's not allowed to download external files on installation of deb packages as per `debian` policies, which makes sense. I guess that if this file grows too big we can somehow group the assets in more than one `deb` package. 

Another option could be having a script that has to be manually triggered by the user or that is triggered upon running any demo for the first time.

rmf_demos_maps/CMakeLists.txt
Yadunund: Say a `Chair` model is used in `office` and `clinic` world, wouldn't this download the model twice? Each one into 
`/install/rmf_demos_maps/share/rmf_demos_maps/maps/office/models/` and `/install/rmf_demos_maps/share/rmf_demos_maps/maps/clinic/models/`

aaronchongth: > I'm ok with this if we have BINARY_BUILD=1 by default (maybe PACKAGE_BUILD is a better name?)

Sounds good, I think you can make the call since you have a better understanding of the semantics in package building.

> the problem with having a non default-set environment variable is that it has to be set before packaging and to do that I have to add a patch on the release repo. I would like to avoid those are much as possible because they are exceptions that have to be treated on a case by case basis and updated when required which make them a good possible point of failure for current and future builds

Yeah, I think it is safe to assume the build is for a source build if the environment variable does not exist or if it exist and its value is 0, while any other case would be building for release.

> Say a Chair model is used in office and clinic world, wouldn't this download the model twice? Each one into
`/install/rmf_demos_maps/share/rmf_demos_maps/maps/office/models/ and /install``/rmf_demos_maps/share/rmf_demos_maps/maps/clinic/models/`

Good catch, I think we will need drop-in replacement for `~/.gazebo/models`, so a common directory of models for all worlds.

Related to the above conversation regarding environment variables, the launch files will also have to check for those environment variables to decide which directory to include into Gazebo's or Ignition's MODEL_PATH.

marcoag: > Say a `Chair` model is used in `office` and `clinic` world, wouldn't this download the model twice? Each one into
> `/install/rmf_demos_maps/share/rmf_demos_maps/maps/office/models/` and `/install/rmf_demos_maps/share/rmf_demos_maps/maps/clinic/models/

This case can be easily avoided by having packages that contain groups of models that are not directly related to demos, the side effect is that every demo will most probably install extra models that they don't need.

Maybe the best option is to try to check for models on demo launch time and directly ask the user to confirm the download of them if they don't exist at that point. This way we avoid the manual download from the user side and having them taking space on the `deb` package.

marcoag: > > I'm ok with this if we have BINARY_BUILD=1 by default (maybe PACKAGE_BUILD is a better name?)
> 
> Sounds good, I think you can make the call since you have a better understanding of the semantics in package building.

I will add these changes for now then.

marcoag:
With the last change models will go to the user's home directory only if `RMF_DOWNLOAD_MODELS_HOME` as been set as environment variable.

aaronchongth (approved):
Thanks for making the fixes! I tested it and it works well. Just a note, before this change, `install/rmf_demos_maps` is 59.4 MB in size, while this change will make `install/rmf_demos_maps` 329.9 MB in size on my machine. If this causes an issue with the binaries, we should revisit this and pursue other methods of releasing maps with models.

PR #69, opened 2021-08-03
remove requried attribute in xml launch file

marcoag:
Signed-off-by: Marco A. Gutierrez <marco@openrobotics.org>

### Fixed bug

Airport launch file throws an error when trying to run the demo:
```
...
launch.invalid_launch_file_error.InvalidLaunchFileError: Caught exception when trying to load file of format [xml]: Unexpected attribute(s) found in `node`: {'required'}
```

Even though the `required` attribute is supported for nodes on python launch files, based on the error above and the documentation found here: https://design.ros2.org/articles/roslaunch_xml.html it seems that it is not available on for `xml` launch files. 

I believe we are only getting this error now on `galactic` due to the changes introduced here: https://github.com/ros2/launch/pull/468.

### Fix applied

`required` attribute has been removed from launch file.


youliangtan (approved):
Thx for fixing this!

PR #76, opened 2021-08-17
add test_ackman level

ddengster:
Signed-off-by: ddengster <ed.fan@osrfoundation.org>

Add a simple level to test nonholonomic pathing.

To test: `ros2 launch rmf_demos test_ackman.launch.xml use_ignition:=1`
To get the ambulance moving (replace `a1` or `c1` with your desired goal points):
`ros2 run rmf_demos_tasks request_ackmann_path -p rmf_demos_maps -n maps/test_ackman/nav_graphs/0.yaml -s a1 -e c1 -i taskid123 -r ambulance`

l1

image

### DIFF
--- a/rmf_demos/launch/test_ackman.launch.xml
+++ b/rmf_demos/launch/test_ackman.launch.xml
@@ -19,19 +19,4 @@
     <arg name="gazebo_version" value="$(var gazebo_version)" />
   </include>
 
-  <!-- TinyRobot fleet adapter and robot state aggregator needed for the TinyRobot slotcar_plugin -->
-  <group>
-    <let name="fleet_name" value="tinyRobot"/>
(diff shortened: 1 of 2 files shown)


luca-della-vedova (approved):
Looks good! Wondering if you think it's possible to make the vehicles less prone to turning in place (the first bend seems a bit unrealistic) but that can be a future TODO for model optimization.

ddengster:
> Looks good! Wondering if you think it's possible to make the vehicles less prone to turning in place (the first bend seems a bit unrealistic) but that can be a future TODO for model optimization.

You can tweak the ambulance's drive speed. I intend to improve the turning code in the `rmf_simulation` repo. But we can put in this level for testing.

FYI my copy of ignition-gazebo5 doesnt have the physics fix so I multiplied the speeds by about a hundred - and it reaches it's waypoint targets correctly.

PR #81, opened 2021-08-31
Remove simulator dependencies from assets

luca-della-vedova:
### Fixed bug

The assets depended on Gazebo, and this made splitting simulation packages (and dependencies) between Gazebo and Ignition not possible.

### Fix applied

Removed the dependency from Gazebo from the assets. Note that to work properly they will still need either the Ignition or the Gazebo plugins to be installed, however since they are just models they can be used and don't technically depend on a simulator, also they will be brought in by a top level simulator specific package anyway (such as `rmf_demos_gz`).


youliangtan (approved):
Yap, having just gz but not ign looks incomplete too. This makes sense.

PR #83, opened 2021-09-02
Finishing request

Yadunund:
* Update `full_control` launch files to support finishing task introduced in https://github.com/open-rmf/rmf_ros2/pull/108.
* Vicinities of several robots were over inflated triggering unnecessary conflicts.


aaronchongth (approved):
Thanks for fixing the inflated values! LGTM

PR #84, opened 2021-09-02
change from holonomic to steering

ddengster:
Signed-off-by: ddengster <ed.fan@osrfoundation.org>

Based off upstream change in https://github.com/open-rmf/rmf_simulation/pull/46


Yadunund (approved):
Thanks for this change

PR #85, opened 2021-09-03
Update changelogs

Yadunund:
Update changelogs and bump pkg versions to `1.3.0`


luca-della-vedova (approved):
Pushed a few minor fixes. LGTM!

PR #86, opened 2021-09-07
Fix Ackermann world topic name, change Ambulance to AmbulanceSlotcar, use robot spawning for test_ackman map

luca-della-vedova:
### Fixed bug

Depends on [rmf_simulation #49](https://github.com/open-rmf/rmf_simulation/pull/49), change the topic name for the ackermann test world to follow the RMF standard, as well as add a bit of documentation for users to launch it.


ddengster (approved):
Thanks for the changes!

ddengster (approved):

Yadunund (changes requested):
I find the name `AmbulanceSlotcar` to be a bit strange since other robot models in `rmf_demos_assets` also include the `slotcar` plugin but are simply called `TinyRobot`, `DeliveryRobot` etc. Also, it seems that the top level folder was renamed to `AmbulanceSlotcar` but the model name in `model.sdf` and `mode.config` is `ambulance` which is a bit confusing.

I understand the need to rename the model as `fuel` may contain other Ambulance models which don't have the plugin included. My recommendation would be to rename to  `EmergencyVehicle` or `HospitalVan`.

README.md
```diff
@@ -226,3 +226,12 @@ $ ros2 launch rmf_demos_gz office_mock_traffic_light.launch.xml
...
 
 Battery recharging is tightly integrated with the new task planner. `ChargeBattery` tasks are optimally injected into a robot's schedule when the robot has insufficient charge to fulfill a series of tasks. Currently we assume each robot in the map has a dedicated charging location as annotated with the `is_charger` option in the traffic editor map.
+
+#### Ackermann test Scenario:
+
+This scenario has a simple demo to show trajectory following for ackermann steering vehicles (such as ambulances). The feature is currently only supported in Ignition, to test the feature:
```
Yadunund, line 232: @luca-della-vedova what makes this only available in `ign`?

aaronchongth (changes requested):
Thanks for updating!

The names of the model `AmbulanceSlotcar` in `model.sdf` and `model.config` are not matching, as well as the names of the `.obj` and `.mtl` file, which causes a clean build to fail.

``` 
[ign-14] [GUI] [Err] [SystemPaths.cc:377] Unable to find file with URI [model://Ambulance/meshes/ambulance.obj]
[ign-14] [GUI] [Err] [SystemPaths.cc:467] Could not resolve file [model://Ambulance/meshes/ambulance.obj]
[ign-14] [GUI] [Err] [MeshManager.cc:172] Unable to find file[model://Ambulance/meshes/ambulance.obj]
[ign-14] [GUI] [Err] [MeshDescriptor.cc:56] Mesh manager can't find mesh named [model://Ambulance/meshes/ambulance.obj]
```

There is currently an issue regarding model downloading that is being tracked, https://github.com/open-rmf/rmf_traffic_editor/issues/386. In the meantime, could you try a clean build, cleaning up `~/.gazebo/models` such that there is only `sun` and `ground_plane` left, and deleting everything in `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics`?

README.md
luca-della-vedova: Good question, in the original implementation of slotcar there was a lot of code specific to the simulator, I believe this might be fixed with the latest [rmf_simulation refactor](https://github.com/open-rmf/rmf_simulation/pull/49), however Gazebo is hanging on startup on my side so not sure what is happening there.

Yadunund:
Also just noticed within the `test_ackman.launch.xml`,  the ```<arg name="use_tpe" value="$(var use_tpe)"/>``` is included within the `common.launch.xml` include. I think it should be under ```<include file="$(find-pkg-share rmf_demos_ign)/simulation.launch.xml">```.

Also in my personal opinion, it seems like a lot of the work to support ackermann vehicles is not fully there yet. I think it would be best to remove the `test_ackman` map and Ambulance model from `main`. We can work on this in a separate branch. Then when everything is finalized (including a proper map for demonstrating ackermann steer vehicles and updated Ambulance model with wheels), we can merge these components into `main` and release as a patch.

luca-della-vedova:
> Also in my personal opinion, it seems like a lot of the work to support ackermann vehicles is not fully there yet. I think it would be best to remove the `test_ackman` map and Ambulance model from `main`. We can work on this in a separate branch. Then when everything is finalized (including a proper map for demonstrating ackermann steer vehicles and updated Ambulance model with wheels), we can merge these components into `main` and release as a patch.

Closing in favor of #88 removing the test world.

PR #88, opened 2021-09-08
Remove test_ackman world

luca-della-vedova:
### Fixed bug

Following discussion in #86, the test_ackman world has quite a few issues, like not working in Gazebo, not actually using RMF, we can keep working on it under the `test/ackermann` branch until it is deemed good enough to be merged into `main`.


Yadunund (approved):
LGTM

Issue #92, opened 2021-09-16
rmf_demos_panel: Download of webpack to location with no permissions


marcoag:
### Required information:

- Operating system and version:
  - Ubuntu Focal
- OpenRMF installation type:
  - binaries
- OpenRMF version or commit hash
  - 1.3.0
- ROS distribution and version:
  - galactic
- ROS installation type:
  - binaries
- Package or library, if applicable:
  - rmf_demos_panel

### Description of the bug

When trying to run a demo on an installed version of Open-RMF `rmf_demos_panel` uses this function to try to download the webpack to a location that is relative to the installation of the package:

https://github.com/open-rmf/rmf_demos/blob/513550af993a9c52a55d06c853576dea4c944a86/rmf_demos_panel/rmf_demos_panel/gui_server.py#L32

when using binary packages the user does not have permissions on the installation directory, therefore this fails:

```
[gui_server-13] Check path  /opt/ros/galactic/lib/python3.8/site-packages/rmf_demos_panel/static/dist/app.bundle.js
[gui_server-13] BundleJS File does not exist, Download it from https://open-rmf.github.io/rmf_demos/v1.0.0/app.bundle.js
[gui_server-13] Traceback (most recent call last):
[gui_server-13]   File "/opt/ros/galactic/lib/rmf_demos_panel/gui_server", line 11, in <module>
[gui_server-13]     load_entry_point('rmf-demos-panel==0.0.0', 'console_scripts', 'gui_server')()
[gui_server-13]   File "/opt/ros/galactic/lib/python3.8/site-packages/rmf_demos_panel/gui_server.py", line 77, in main
[gui_server-13]     download_webpack()
[gui_server-13]   File "/opt/ros/galactic/lib/python3.8/site-packages/rmf_demos_panel/gui_server.py", line 51, in download_webpack
[gui_server-13]     open(bundle_path, 'wb').write(response.content)
[gui_server-13] PermissionError: [Errno 13] Permission denied: '/opt/ros/galactic/lib/python3.8/site-packages/rmf_demos_panel/static/dist/app.bundle.js'
```
Alternative solutions to the problem
- Install the webpack somewhere in some config directory in the user's home directory.
- Find alternative ways to distribute the webpack. can this be installed through system packages? any other way?

### Steps to reproduce the bug

1. On a new Ubuntu focal machine install `ros-galactic-rmf-demos-gz`.
[...]

PR #93, opened 2021-09-21
migrate rmf_panel js component to a new repo

youliangtan:
## Brief Description

This is to address https://github.com/open-rmf/rmf_demos/issues/92.

After discussion, since the react source code is located within `rmf_demos_panel` static dir, it doesn't make sense to download the compiled version from the web. However, if we wish to compile the webpack locally, this will introduce a dependency to `npm` (and other node modules), which is HUGE by itself.

Thus a new repo is created to host the front end component of `rmf_demos_panel: https://github.com/open-rmf/rmf-panel-js/pull/1

This in turn will simplify the setup process. All documentation are updated accordingly.

## Run Rmf Demos

Similar steps to launch the demo world as before, instead of accessing the dashboard from a local GUI server (localhost:5000), now the user can straightaway access the link here: https://open-rmf.github.io/rmf-panel-js/  (which is hosted by github page)

Note that the `simple-api-server` that act as the interface between rmf_core and web is still located within `rmf_demos_panel`. port number is now configurable and change to `8083` to prevent conflict with the commonly used `8080`


marcoag (approved):

marcoag:
Great work! awesome stuff!

PR #105, opened 2021-11-03
Fix incorrect lift_watchdog srv name 

youliangtan:
Mainly is just a simple one-liner fix on the incorrect `experimental_lift_watchdog_service` name in the `hotel.world`. This can be easily run by: `ros2 launch rmf_demos_gz hotel.launch.xml enable_experimental_lift_watchdog:=1`

Sneaked in some readme updates on how to use lift_watchdog, and other additional features of rmf.

### DIFF
--- a/rmf_demos/launch/hotel.launch.xml
+++ b/rmf_demos/launch/hotel.launch.xml
@@ -19,7 +19,7 @@
           exec="experimental_lift_watchdog"
           name="experimental_lift_watchdog"
           output="both" />
-    <set_env name="EXPT_LIFT_WATCHDOG_SRV" value="experimental_lift_watchdog_service"/>
+    <set_env name="EXPT_LIFT_WATCHDOG_SRV" value="experimental_lift_watchdog"/>
   </group>
   <group unless="$(var enable_experimental_lift_watchdog)">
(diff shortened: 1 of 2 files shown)


rmf_demos/launch/hotel.launch.xml
```diff
@@ -19,7 +19,7 @@
           exec="experimental_lift_watchdog"
           name="experimental_lift_watchdog"
           output="both" />
-    <set_env name="EXPT_LIFT_WATCHDOG_SRV" value="experimental_lift_watchdog_service"/>
+    <set_env name="EXPT_LIFT_WATCHDOG_SRV" value="experimental_lift_watchdog"/>
```
mxgrey, line 22: Silly question.. Where does this environment variable actually get used? I can't find any reference to it in `rmf_demos` or in `rmf_ros2`.

youliangtan: This is used in the launch file  [here](https://github.com/open-rmf/rmf_demos/blob/6de2a7655f2deb2c62f0c7890fd7608a97cdae2e/rmf_demos/launch/hotel.launch.xml#L70). The reason of using ENV is because `launch.xml `'s `<arg>` element doesn't support the "if/unless"  condition.

mxgrey (approved):

PR #110, opened 2021-11-12
fix typo in readme for CLI task dispatch

codebot:
Fix typo so the CLI task dispatch example works in the README for the hotel demo.

Signed-off-by: Morgan Quigley <morgan@osrfoundation.org>


Yadunund (approved):
Thanks for the fix!

PR #122, opened 2022-02-08
Helpful scripts for submitting composite tasks

Yadunund:
This PR introduces additional scripts in `rmf_demos_tasks` to serve as examples of how to submit task requests defined using the flexible task definitions.

New scripts: 
- dispatch_patrol
- dispatch_clean
- dispatch_action
- dispatch_delivery
- cancel_task

Most of these scripts also accept `-robot -fleet` as input args, if the user would like to directly assign a task to a specific robot in a fleet. `dispatch_patrol` and `dispatch_delivery` support multiple places/pickup-dropoff points. `dispatch_action` is also in a more generic form.

For additional details, refer the readme [here](https://github.com/open-rmf/rmf_demos/blob/redesign_v2/rmf_demos_tasks/README.md)


Yadunund:
@mxgrey should we consider merging this in so that users can look at how composite tasks can be submitted?

youliangtan:
@Yadunund Is this ready to review? i can help to test it and review it.

Yadunund:
@Yadunund I think it's ready

youliangtan:
I just cleaned up all the utils scripts in `rmf_demos_tasks`, and update the docs respectively. Previously we have many "kinda duplicated" scripts, and it can get a little bit confusing to users. To sum it up, we will now have 5 types of scripts, compatible with task v2:
- dispatch_patrol
- dispatch_clean
- dispatch_action
- dispatch_delivery
- cancel_task

The `dispatch_xxx` scripts can also accept `-robot -fleet` as input args, to support direct `robot_task_request` assignment

For detailed info, can refer to the readme [here](https://github.com/open-rmf/rmf_demos/blob/redesign_v2/rmf_demos_tasks/README.md)

Feel free to test those scenarios out again. If no issue with this, hopefully we can merge this to `main` as soon as possible.

rmf_demos_tasks/rmf_demos_tasks/dispatch_action.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
```
Yadunund, line 1: To me the point of the scripts in this package is to act as a convenience tool for issuing a particular task. This script however aims to be very generic while at the same time has a couple of "gotchas". For example, I can submit a list of `start` locations and the robot will be guided to each of these but it will always perform the same action at each location. We do not accept a corresponding list of actions. It also make an implicit assumption on the `use_took_sink` parameter depending on the action string. We're also assuming that the action description is a string that is serialized from a json adhering to the [PerformAction](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/schemas/event_description__perform_action.json) schema.

For these reasons I'm inclined to leaving it out. But if you feel it has been helpful then I guess we can have it. But I'd recommend updating the readme to describe the parameters in more detail and also change the behaviour to only accept a single start and matching action.

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2021 Open Source Robotics Foundation, Inc.
+# Copyright 2022 Open Source Robotics Foundation, Inc.
```
Yadunund, line 3: Same as previous comment

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
```diff
@@ -50,60 +61,99 @@ def __init__(self, argv=sys.argv):
...
+            wl = workcell_list.split(',')
+            return {
+                    "place": wl[0],
+                    "handler": wl[1],
+                    "payload": [{"sku": wl[2],
+                                "quantity": int(wl[3])}]
+                    }
+
+        # Use standard delivery task type
+        if len(self.args.pickups) == 1 and len(self.args.dropoffs) == 1:
```
Yadunund, line 111: You can just have this as always `compose`.

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
```diff
@@ -50,60 +61,99 @@ def __init__(self, argv=sys.argv):
...
+        # Set task request start time
+        now = self.get_clock().now().to_msg()
+        now.sec = now.sec + self.args.start_time
+        start_time = now.sec * 1000 + round(now.nanosec/10**6)
+        request["unix_millis_earliest_start_time"] = start_time
+
+        def __create_workcell_desc(workcell_list):
+            wl = workcell_list.split(',')
+            return {
+                    "place": wl[0],
```
Yadunund, line 104: This will throw a runtime error if the user runs the script with fewer comma separated parameters that what this is expecting.

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
```diff
@@ -16,30 +16,41 @@
...
         parser = argparse.ArgumentParser()
-        parser.add_argument('-p', '--pickup', required=True,
-                            type=str, help='Start waypoint')
-        parser.add_argument('-pd', '--pickup_dispenser', required=True,
-                            type=str, help='Pickup dispenser name')
-        parser.add_argument('-d', '--dropoff', required=True,
-                            type=str, help='Finish waypoint')
-        parser.add_argument('-di', '--dropoff_ingestor', required=True,
-                            type=str, help='Dropoff ingestor name')
+        parser.add_argument('-p', '--pickups', required=True,
```
Yadunund, line 42: It's great that we're trying to support multiple pickup/drop-offs. But I find packing the pickup and dropoff params into a string which is then split at `,` to be very un-intuitive. Suggest to leave the original args ( -p, -pd, -d, -di) and add any others that you may require (sku, qty etc). Then have these args accept arrays via `nargs='+'`. Within the script you check if the lengths of all the arrays are equal and if so add pickup and dropoff descriptions with values at corresponding indices. If there is a mismatch in the array lengths, we can print an error message for the user.

rmf_demos_tasks/rmf_demos_tasks/teleop_action.py
```diff
@@ -0,0 +1,78 @@
...
+        msg = PathRequest()
+        msg.fleet_name = self.args.fleet
+        msg.robot_name = self.args.robot
+        msg.task_id = str(uuid.uuid1())
+        for p in self.args.points:
+            pts = p.split(',')
+            loc = Location()
+            loc.x = float(pts[0])
+            loc.y = float(pts[1])
+            loc.yaw = 0.0
```
Yadunund, line 61: why not get the yaw from `p`?

rmf_demos_tasks/rmf_demos_tasks/teleop_action.py
```diff
@@ -0,0 +1,78 @@
...
+          PathRequest, 'robot_path_requests', 1)
+
+        msg = PathRequest()
+        msg.fleet_name = self.args.fleet
+        msg.robot_name = self.args.robot
+        msg.task_id = str(uuid.uuid1())
+        for p in self.args.points:
+            pts = p.split(',')
+            loc = Location()
+            loc.x = float(pts[0])
```
Yadunund, line 59: We should check for out of bounds before directly accessing via indices.

rmf_demos_tasks/README.md
```diff
@@ -2,6 +2,85 @@
...
+  ros2 run rmf_demos_tasks dispatch_action -id patrol.dispatch-0
+  ```
+
+## Additional Scripts
+
+- `teleop_action`
+
+  This script demonstrates sending a teleop action with a particular set of coordinates `-p` you would like your robot to follow in a given map `-m`. To test it, launch the Office world and run the above `dispatch_action` task to bring `tinyRobot1` to `coe`. Then,
+  ```
+  ros2 run rmf_demos_tasks teleop_action -F tinyRobot -R tinyRobot1 -p 5.3,-4.9 3.3,-8.9 5.3,-9.9 7.3,-6.1 5.3,-4.9 -m L1
```
Yadunund, line 78: Suggest to create a launch file, `office_teleop_robot.launch.xml`  inside rmf_demos/office that will invoke this script with the correct set of -p values.

rmf_demos_tasks/rmf_demos_tasks/teleop_action.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
Yadunund, line 1: Let's rename this file to `teleop_robot.py`. The `teleop_action` is a bit ambiguous given we have `dispatch_teleop` and `dispatch_action` scripts.

rmf_demos_tasks/setup.py
```diff
@@ -26,11 +26,13 @@
         'console_scripts': [
           'request_loop = rmf_demos_tasks.request_loop:main',
           'request_lift = rmf_demos_tasks.request_lift:main',
-          'dispatch_loop = rmf_demos_tasks.dispatch_loop:main',
```
Yadunund, line 29: Let's not remove `dispatch_loop` yet as users may still be using the script to submit loop requests. We could update the script to print a deprecation warning and inform users to use `dispatch_patrol` instead. We can consider removing the script for `humble` release where "breaking" things would be acceptable.

.github/workflows/build.yaml
```diff
@@ -5,20 +5,18 @@ jobs:
   ros2:
     runs-on: ubuntu-20.04
     container:
-      image: docker://ros:foxy-ros-base-focal
+      image: docker://ros:galactic-ros-base-focal
```
Yadunund, line 8: Can we open a separate PR to merge the CI fixes ASAP? I don't think it should be part of this PR.

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
youliangtan: I have changed the format of the script. An example is shown in the readme. 
  ```bash
  ros2 run rmf_demos_tasks dispatch_delivery \
  -p pantry pantry \
  -ph coke_dispenser coke_dispenser_2 \
  -d hardware_2 coe \
  -dh coke_ingestor coke_ingestor_2 \
  -pp coke,1 coke,1 \
  -dp coke,1 coke,1 \
  --use_sim_time
  ```

Within the script, it will conduct assertion checks.

rmf_demos_tasks/rmf_demos_tasks/dispatch_delivery.py
youliangtan: I was also pondering about this. However, the reason is to provide an example of using [task_description__delivery](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/schemas/task_description__delivery.json), instead of composing a series of events.

rmf_demos_tasks/rmf_demos_tasks/dispatch_action.py
youliangtan: Understand. I made a slight update on the readme, to clarify that only 1 type of custom action is accepted. Also, change the script to make `--use_took_sink` as an input arg, similar to `--use_sim_time`. Note that the `action description` is actually an optional description in perform action ([here](https://github.com/open-rmf/rmf_ros2/blob/91fe9b61a901f50a81e6b1b20700ee1615964a22/rmf_fleet_adapter/schemas/event_description__perform_action.json#L12-L13)).

This generic form of `dispatch_action` script is quite handy for testing out simple custom perfomactions. An example of 
 new composed cleaning action task:
 ```
 ros2 run rmf_demos_tasks dispatch_action -s lounge -a clean -ad '{ "clean_task_name": "clean_lounge" }' --use_took_sink
 ```

Yadunund:
Thanks for the changes @youliangtan. 
Github won't let me approve this PR since I'm the original author. But we can merge this one once someone else does so.

mxgrey (approved):

PR #123, opened 2022-02-11
redesign_v2: update rmf-demos-panel

youliangtan:
For task redesign_v2: `rmf-panel` will not work when testing with the new task `redesign_v2`.  https://github.com/open-rmf/rmf_ros2/pull/168

For more serious users of web-based rmf, it is recommended to use `rmf-web`. However, `rmf-panel` is still handy for users to quickly test out `rmf_demos` with a simple web dashboard, so fixes are applied here.

In the `simple-api-server`
- update previous submit task srv call to new `task_api_requests`
- To listen to `task_state_updates`, a WebSocket server is setup in `simple-api-server` (default port 7878) 
- task cancellation `task_api_requests` is still not available. Will add it when the feature is fully tested.

## To Test

Similar to how we launch the world as before:

```
ros2 launch rmf_demos_gz office.launch.xml server_uri:="ws://localhost:7878"
```

Then open the same frontend dashboard to request and view tasks. https://open-rmf.github.io/rmf-panel-js/

p/s: if we wish to consolidate all PR to `redesign_v2` for the major merge, I can rebase this pr to `redesign_v2`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,15 @@ There are two main modes of submitting tasks to RMF via the Panel:
 1. Submit a Task: Used to submit a single task.
 2. Submit a List of Tasks: Used to submit a batch of tasks. A `.json` file containing a list of tasks may be loaded via the `Choose file` button. Some example files are found in `rmf_demos_panel/task_lists`.
 
+**LATEST UPDATE: Display Task States**
+
+`task_state_uptates` are now published via websocket. To display task states on `rmf-panel`, specify `server_uri:="ws://localhost:7878"` during ros2 launch. Example:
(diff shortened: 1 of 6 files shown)


mxgrey (approved):
I'll go ahead and merge this since it at least fixes the crash during startup.

However I think it needs some more work, because when I tried it out it was unable to display task states. It also eventually crashed with this error:

```
[simple_api_server-12]     Task Status: {json.dumps(task_status.json)}")
[simple_api_server-12] AttributeError: 'Response' object has no attribute 'json'
```

PR #128, opened 2022-03-07
Add parameter to force Ignition Version

luca-della-vedova:
### Implemented feature

This PR adds a parameter to force the Ignition Gazebo version when running `rmf_demos_ign` launch files.

The plugins in [rmf_simulation](https://github.com/open-rmf/rmf_simulation/) are compiled against a specific version of Ignition Gazebo and if a different version of Ignition than the one used to compile the plugins was to be ran, the simulation would crash.
By default, the `ign gazebo` CLI runs the most recent version of Gazebo, hence if a user was to install a newer version than the one supported demos would stop working.

This PR introduces a `ignition_version` parameter to `rmf_demos_ign` that defaults to the version plugins are built against. Since it specifies Fortress it should be merged together with https://github.com/open-rmf/rmf_simulation/pull/70 to complete our migration. Users who wish to use different Ignition version can change the parameter value by doing, for example:

`ros2 launch rmf_demos_ign office.launch.xml ignition_version:=6`

### DIFF
--- a/rmf_demos_ign/launch/airport_terminal.launch.xml
+++ b/rmf_demos_ign/launch/airport_terminal.launch.xml
@@ -4,6 +4,7 @@
   <arg name="use_sim_time" default="true"/>
   <arg name="failover_mode" default="false"/>
   <arg name="use_tpe" default="false"/>
+  <arg name="ignition_version" default='6'/>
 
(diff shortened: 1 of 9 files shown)


aaronchongth (approved):
Thanks for adding this support!

PR #130, opened 2022-03-21
Added instructions to change `server_uri` when using `rmf-web` dashboard

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

## Documentation

Adding the instruction to target the `server_uri` launch parameter when using the web dashboard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ A near-term roadmap of the RMF project can be found in the user manual [here](ht
 
 Full web application of RMF: [rmf-web](https://github.com/open-rmf/rmf-web).
 
+In order to interact with the default configuration of the web application, the `server_uri` launch parameter will need to be changed to `ws://localhost:8001`, for example,
+
+```bash
+ros2 launch rmf_demos_ign office.launch.xml server_uri:="ws://localhost:8001"
+```
+
 ## Demo Worlds
 
 * [Hotel World](#Hotel-World)


Yadunund:
Not sure if I'm missing something but I don't see the `server_uri` argument defined in any of the launch files or passed as a Param into the adapters?

mayman99:
@Yadunund the `server_uri` is the `fleet_adapter` launch file.

mayman99 (approved):

Yadunund:
@mayman99 i'm aware but the argument is only defined in the [launch file](https://github.com/open-rmf/rmf_ros2/blob/7ed2196c37edc444ba065847b399c73891f4d681/rmf_fleet_adapter/launch/fleet_adapter.launch.xml#L43) in the `rmf_fleet_adapter` library 

But all these adapters that control the robots in the simulation here have [these](https://github.com/open-rmf/rmf_demos/blob/65c45a1f52c5d909381373b35df4146d4069e7f5/rmf_demos/launch/include/adapters/tinyRobot_adapter.launch.xml#L12) launch files in the `rmf_demos` package that include the above launch file and overwrite the relevant parameters. But as you can see, the `server_uri` parameter is not explicitly set from any argument that can be overwritten by the demo launch files.

My issue is with how the `server_uri` parameter values makes it all the way from [rmf_demos_ign/office.launch.xml](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_ign/launch/office.launch.xml) -> [rmf_demos/offcie.launch.xml](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos/launch/office.launch.xml) -> [tinyRobot_adapter](https://github.com/open-rmf/rmf_demos/blob/65c45a1f52c5d909381373b35df4146d4069e7f5/rmf_demos/launch/office.launch.xml#L18-L22) without any of these launch files defining an `  <arg name="server_uri"/>` element? Do undefined parameters just automatically get passed as a ROS 2 parameter to every node in the launch file?

mayman99:
> My issue is with how the server_uri parameter values makes it all the way from [rmf_demos_ign/office.launch.xml](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_ign/launch/office.launch.xml) -> [rmf_demos/offcie.launch.xml](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos/launch/office.launch.xml) -> [tinyRobot_adapter](https://github.com/open-rmf/rmf_demos/blob/65c45a1f52c5d909381373b35df4146d4069e7f5/rmf_demos/launch/office.launch.xml#L18-L22) without any of these launch files defining an <arg name="server_uri"/> element? Do undefined parameters just automatically get passed as a ROS 2 parameter to every node in the launch file?

Yeah I think this is the case. I remember it actually took me a while to figure out that `server_uri` parameter when I first started using `rmf`, maybe we could refactor it's usage to be a bit more obvious, e.g. move it to higher level launch files, or chain it using `arg` elements in the launch files.

aaronchongth:
Yup! We had this discussion on slack as well, turns out this might be an intended behavior. I think I will go ahead and merge this for now

PR #156, opened 2022-09-01
Minor fixes for demo fleet adapter

xiyuoh:
Fixes small bugs including:
- formatting
- `slotcar` deletes the first waypoint when it receives a `PathRequest`, so it doesn't receive the target waypoint proper during docking. This PR appends the robot's current waypoint to the dock `PathRequest` (which was done similarly in `navigate()`)


luca-della-vedova (approved):
Thanks for the fixes!

PR #157, opened 2022-09-08
Async behavior fixes

mxgrey:
This PR introduces changes to the `rmf_demos_fleet_adapter` API and implementation that help eliminate the risk of certain race conditions that were able to happen in cases where the fleet adapter / fleet manager / slotcar pipeline falls dramatically out of sync. The issues encountered were relatively rare, require long simulation run times and/or running the simulation at higher than real time speeds, but as the rare async issues accumulate they result in the traffic and task systems becoming dysfunctional.

This PR depends on these others:
* https://github.com/open-rmf/rmf_internal_msgs/pull/45
* https://github.com/open-rmf/rmf_traffic/pull/86
* https://github.com/open-rmf/rmf_ros2/pull/228


luca-della-vedova:
Note that this branch is based on `luca/full_control_battle_royale` that made the `battle_royale` scenario use the full control fleet adapter instead of the previous traffic light.
I added a commit (https://github.com/open-rmf/rmf_demos/pull/157/commits/b26204aa54606dcf47b01f10b4b220fe3648384c) to make it configurable, users can use the `use_traffic_light` launch parameter, default to false, to choose whether the demo should run on full control or traffic light.

mxgrey:
Thanks for making the battle royale configurable. It turned out to be an excellent test for the full control adapter, so I think it's good to keep full control as the default.

xiyuoh:
Tested on hotel world with multiple tasks that cross paths, the fixes definitely improves performance. Previously the yellow markers move ahead without the robots sometimes, i'm guessing it is due to the duration estimate problem. With the fixes this doesn't really happen anymore and there seems to be no more ridiculous delay error messages.

When left running long enough (RTF 0.8), the robots may get weird trajectories out of the nav graphs. I'm not entirely sure if this issue should be addressed here, but we encountered it before with the legacy fleet adapter in simpler scenarios.

image

mxgrey:
> When left running long enough (RTF 0.8), the robots may get weird trajectories out of the nav graphs.

I saw this happening once in a while in the battle royale, although not as extreme as what you've pictured.

This kind of behavior would come from a combination of two factors:
* The fleet adapter replans for the robot either because the robot [reports an obstacle](https://github.com/open-rmf/rmf_demos/blob/b26204aa54606dcf47b01f10b4b220fe3648384c/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L321-L322) or because [a negotiation failed](https://github.com/open-rmf/rmf_ros2/blob/83c4df0d48876455163ac1e23f65a21d451bc85d/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L1349-L1375).
* The [robot position updating](https://github.com/open-rmf/rmf_demos/blob/b26204aa54606dcf47b01f10b4b220fe3648384c/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L515-L554) is flawed.

I believe both need to be happening together in order to get this effect. I think making sure the position updating is being done correctly would be the key to eliminating this misbehavior.

I'm reluctant to suggest this but we may want to start with a test that replaces that whole position update section with simply [`update_lost_position`](https://github.com/open-rmf/rmf_demos/blob/b26204aa54606dcf47b01f10b4b220fe3648384c/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L550-L554). One issue that I just spotted is that we never seem to update `self.map_name` with the current map name. Each time we [navigate to a new waypoint](https://github.com/open-rmf/rmf_demos/blob/b26204aa54606dcf47b01f10b4b220fe3648384c/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L290-L318) we should update `self.map_name` with the map name of the new target waypoint.

rmf_demos/launch/common.launch.xml
```diff
@@ -15,10 +15,6 @@
   <node pkg="rmf_traffic_ros2" exec="rmf_traffic_schedule" output="both" name="rmf_traffic_schedule_primary">
     <param name="use_sim_time" value="$(var use_sim_time)"/>
   </node>
-  <!-- Traffic Schedule monitor and replacement node -->
-  <node pkg="rmf_traffic_ros2" exec="rmf_traffic_schedule_monitor" output="both" name="rmf_traffic_schedule_backup">
-    <param name="use_sim_time" value="$(var use_sim_time)"/>
-  </node>
```
luca-della-vedova, line 21: Admittedly I don't know much about how this node works but any reason why it was removed? Was it due to the liveliness callback getting triggered spuriously once in a while and initiating fallback behavior?

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -112,6 +112,7 @@ def initialize_fleet(config_yaml, nav_graph_path, node, use_sim_time):
         datetime.timedelta(seconds=1.0/fleet_state_update_frequency))
     # Account for battery drain
     drain_battery = fleet_config['account_for_battery_drain']
+    lane_merge_distance = fleet_config.get('lane_merge_distance', 1.0)
```
xiyuoh, line 115: We may want to consider decreasing the default `lane_merge_distance`, I believe it is what's causing some of the bad route generation. When filtering waypoints, for routes that include a perpendicular (or close to perpendicular) trajectory, if the robot's pose is within the `lane_merge_distance` of the corner waypoint, [this filter](https://github.com/open-rmf/rmf_demos/blob/async_behavior_fixes/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L649-L652) will delete the waypoints leading to the target, resulting in the "shortcut" path. This tends to occur when the robot is provided with a new path after a negotiation, and it happens to be positioned near the corner waypoint. Tried running a few rounds of simulation with a lane merge distance of 0.1, seems to work fine so far.

Alternatively we could add `lane_merge_distance` to the full control adapters' configs if we want to keep the default at 1.0.

mxgrey: Yeah you're right, 1m is unreasonable for our slotcars. They stay very close to the lanes anyway.

mxgrey: https://github.com/open-rmf/rmf_demos/pull/157/commits/ef08ed498a79f0a5d63eae2be39a6ec46bf6305b

luca-della-vedova (approved):

PR #158, opened 2022-09-21
EasyFullControl integration with rmf_demos

xiyuoh:
This PR depends on https://github.com/open-rmf/rmf_ros2/pull/235.

It demonstrates how you can create a fleet adapter with the EasyFullControl API and its python bindings:
- A fleet adapter is created using `EasyFullControl::make`, and appropriate callbacks such as get_position, navigate, dock, etc. modified from RobotClientAPI are passed to the adapter via `EasyFullControl::add_robot` for each robot that should be added to the fleet. These callbacks are used by the internal fleet adapter code to receive updates from and send commands to the fleet robots.
- An `action` field is added to the Office world's tinyRobot config file to support teleop actions as part of the EasyFullControl API.
- An argument `easy_fleet` is also added to the fleet adapter launch file to easily switch between the current `rmf_demos_fleet_adapter` and the EasyFullControl adapter implementations.

To test, you may run any full control demo world with an `easy_fleet:=1` flag. For example:
```
ros2 launch rmf_demos_gz_classic office.launch.xml easy_fleet:=1
```


Yadunund:
I don't think we need to keep the previous fleet adapter implementation since this was copied into the implementation of EasyFullControl?

xiyuoh:
@Yadunund do we need to keep the python implementation in case some users want to modify the internal logic for their own application?

mxgrey (approved):

PR #160, opened 2022-09-21
Fix teleop action

xiyuoh:
This PR fixes a bug where the robot doesn't carry out a teleop action sent externally via the `teleop_robot` script due to:
- Incompatible QoS when publishing to `/robot_path_requests`
- Fleet manager checks whether the `/robot_state` and internally stored task_id's match and republishes the last PathRequest if they don't. This interrupts the robot when carrying out the teleop action, which carries a different task_id since it was sent directly to slotcar. 

This fix checks that the `/robot_state` task_id is a valid uuid and if so, it assumes that the robot is carrying out a teleop action (instead of the robot state update being out of date). It also checks whether the last PathRequest has been completed successfully before republishing it to slotcar.

### DIFF
--- a/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotClientAPI.py
+++ b/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotClientAPI.py
@@ -187,6 +187,24 @@ class RobotAPI:
             return response['data'].get('replan', False)
         return False
 
+    def toggle_action(self, robot_name: str, toggle: bool):
+        '''Request to toggle the robot's mode_teleop parameter.
+           Return True if the toggle request is successful'''
+        url = self.prefix +\
+            f"/open-rmf/rmf_demos_fm/toggle_action?robot_name={robot_name}"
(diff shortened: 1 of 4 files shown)


Yadunund:
>Fleet manager checks whether the /robot_state and internally stored task_id's match and republishes the last PathRequest if they don't.

Why is this the default behavior? It sounds like the problem would not occur if the fleet manager does not republish the previous path request when there is a discrepancy?

>This fix checks that the /robot_state task_id is a valid uuid and if so, it assumes that the robot is carrying out a teleop action

I'm not a fan of having such assumptions in the logic.
Why not add a boolean parameter to the `State` class of a robot, `mode_teleop`, that can be toggled through an API endpoint. The perform action implementation can be updated to additionally call this API at the start. Internally in the fleet manager, you can ignore discrepancies in task id if `mode_teleop` is `true`.

xiyuoh:
@Yadunund Done, added a `toggle_action` endpoint which is triggered by the fleet adapter when an action waypoint index is present

Yadunund (approved):

PR #161, opened 2022-09-22
[CI] Reusable CI

orensbruli:
Signed-off-by: Esteban Martinena <orensbruli@gmail.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,53 +1,24 @@
 name: build
 on:
-  pull_request:
+  pull_request: 
+  schedule:
+    - cron: '55 0 * * *'
+  workflow_dispatch: 
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        docker_image: ['ros:humble-ros-base']
-    container:
-      image: ${{ matrix.docker_image }}
-    steps:
-    - name: non-ros-deps
-      run: |
-        sudo apt update
-        sudo apt install -y wget
-        sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -1,53 +1,15 @@
...
-        export CXX=clang++
-        export CX=clang
-        colcon build --mixin lld --cmake-args -DNO_DOWNLOAD_MODELS=True
+  build_and_test:
+    name: rmf_demos
+    uses: open-rmf/rmf_ci_templates/.github/workflows/reusable_build.yaml@main
+    with:
+      # NOTE: Avoid adding comments in the packages lines, this can break some of the called scripts in github actions
+      packages: |
+        rmf_demos
```
luca-della-vedova, line 15: Should this contain all the packages in the repo rather than only `rmf_demos`?

orensbruli: You are absolutely right. Thank you. Fixing

luca-della-vedova:
Two main comments:

Galactic and Foxy are not supported on `main` anymore, since this package depends on `rmf_simulation` which itself doesn't support them anymore. We should add the matrix to make sure we only test `main` under Jammy + Humble / Rolling.

I see linter errors in the `rmf_demos_fleet_adapter` that I can also reproduce through `colcon test` locally, not sure whether to include them in this PR or not. I also noticed that they were not caught by the `style` action, it seems to be a less strict constraint than flake8 / pep257 that are part of tests.
Also there are multiple Python packages (i.e. `rmf_demos_tasks`, `rmf_demos_panel`), but only one implements the linting tests (`rmf_demos_fleet_adapter`).

I would suggest unifying by either:

* Remove the flake8 / pep257 tests from `rmf_demos_fleet_adapter`
* Add flake8 / pep257 to all remaining Python packages and remove the `style` action, only relying on `colcon test` for style.

My personal preference would be for the second, to make this package in line with the rest of the RMF packages that have no style action but uncrustify / style as part of their tests, but I suspect introducing linting to these packages that never had it in the first place would bring quite a huge diff.

Yadunund:
Closed by #208

PR #170, opened 2022-12-01
fix broken build by vendoring pkgs

koonpeng:
### Fixed bug

This is based on https://github.com/PranayShirodkar/rmf_demos/tree/feature/add_speed_limits however the same problems exist on the main branch. `rmf_demos_fleet_adapter` uses a number of python packages not available in ros humble, these packages are not declared in `package.xml` so it will not work without manually installing those packages.

There will be errors like:

```
ModuleNotFoundError: No module named 'fastapi'
```

### Fix applied

Most of the packages are available in ubuntu repos but they are outdated, declaring them in `package.xml` will cause an incompatible version to be installed. The fix applied is to instead vendor these packages.


luca-della-vedova:
As of Ubuntu 24.04 the package mismatch issue of fastapi has been addressed, and with #211 they are specified as dependencies and installed from apt so this should no longer be needed.

PR #176, opened 2023-03-30
Fix conversion from minutes to milliseconds in rmf_demos_panel

mxgrey:
It was brought to our attention that the conversion from the old `rmf_demos_panel` format to the new `rmf_api_msgs` format had an error where minutes were not being correctly converted into milliseconds (instead they were being converted into seconds).

This PR fixes that, which is important for backwards compatibility of the `rmf_demos_panel`.


aaronchongth (approved):
Thanks for the fix!

PR #180, opened 2023-05-09
add requester name to finishing task

aaronchongth:
### Implemented feature

Using new python API introduced in https://github.com/open-rmf/rmf_ros2/pull/267, to provide a requester name for the finishing task factories.


aaronchongth:
Closing as the node name is passed from within the fleet update handle via https://github.com/open-rmf/rmf_ros2/pull/267

PR #183, opened 2023-05-25
Sync main with rolling

Yadunund:
~~We have been bumping versions on rolling and humble without porting to `main`. This PR syncs main back.~~

Noticed that there was a PR https://github.com/open-rmf/rmf_demos/pull/167 in the past to port changes from `rolling` into `main` but this PR was squash merged instead of a commit merge. As a results the version bumps and more importantly the tags for the release did not make to `main` since a tag is associated with a commit.

So although this PR does not have a code diff, we should merge it with a merge commit to ensure the commits with the version tags become present in `main`.


aaronchongth (approved):
thanks!

Yadunund:
Looks like I have to disable the branch protection rule to ensure commit history is linear if I want to do a merge commit. If we do change history in a non-linear manner by merging this, any idea what the worst case scenario is for downstream users?Would they need to re-clone this repo if they want to pull newer commits?

aaronchongth:
I might be wrong, but afaik, this won't affect downstream users. Any sort of conflicts will arise before we can merge it, which after resolving, should still make all `git pull`s work. @marcoag :bow: can you confirm this?

Only when we want to change the non-linear history back to linear history (where a force push is required), will users need to reclone, I believe.

Yadunund:
That makes sense. After some thought I think I'm with closing this PR to avoid any future issues. Originally I had planned to delete the rolling branches but I think we can just keep them so that the tags will still exist in a valid branch. Will just bump version on main and tag the new version before doing the source release.

PR #188, opened 2023-07-12
Reactive easy full control

mxgrey:
This PR updates #158 to be compatible with https://github.com/open-rmf/rmf_ros2/pull/286


rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -42,325 +38,15 @@
 from rclpy.qos import qos_profile_system_default
 
 from .RobotCommandHandle import RobotCommandHandle
```
luca-della-vedova, line 40: This crashes at startup with a module not found error because the file does not exist anymore, it should probably be removed

luca-della-vedova:
I noticed a quite reproducible (most times, but not 100% of the times) jump in the schedule after a negotiation. I'm not sure what implications it might have but it seems a bit odd.

I can reproduce it best at startup in the battle royale scenario:

```
ros2 launch rmf_demos_gz_classic battle_royale.launch.xml
ros2 launch rmf_demos battle_go.launch.xml
```

An example video is below, robot with the schedule jump is in the top left. In all my tests it's either TinyRobotA or TinyRobotB in the top left of the map.

[Screencast from 2023-07-13 10-59-19.webm](https://github.com/open-rmf/rmf_demos/assets/9111558/12b1ca66-2b14-46e9-a4d5-3550c36e8f64)

xiyuoh:
After some brief testing, I noticed that when a cleanerbot's cleaning path gets in the way of a patrol robot's route, the robots get stuck in a deadlock without negotiation. Not 100% sure but it might be because trajectory scheduling is not implemented for the cleaning perform action ([example](https://github.com/open-rmf/rmf_demos/pull/162/files#diff-d4c7f8f4b9feb8dfdec5c912cb2754a7f139902b138b2d2bdf758cef701f257eR519-R531)) .

https://github.com/open-rmf/rmf_demos/assets/30265743/d2c436b4-9b0c-4d19-9a03-6faef0aa4f8d

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
mxgrey: Whoops, fixed here: [ad48b8a](https://github.com/open-rmf/rmf_demos/pull/188/commits/ad48b8af12988576bcc8af8072dc1955f77a4eba)

mxgrey:
> it might be because trajectory scheduling is not implemented for the cleaning perform action

Strange, [it is implemented](https://github.com/open-rmf/rmf_demos/blob/ad48b8af12988576bcc8af8072dc1955f77a4eba/rmf_demos/config/hotel/cleanerBotA_config.yaml#L48-L73). I might suggest trying a clean build with the `reactive_easy_full_control` branches for both `rmf_ros2` and `rmf_demos`. Here's a video of what it should look like when the robot is cleaning:

https://github.com/open-rmf/rmf_demos/assets/1307217/ec25c006-1831-4f03-9066-395d5d93cbc1

Maybe try a test where you just send the cleaning bot with nothing else running and see if the cleaning trajectory shows up on the schedule.

xiyuoh:
Ah yes you're right, a clean build and the trajectories are back up! I can still reproduce the same situation though. Haven't figured out a potential cause, leaving a screen recording with terminal logs here first.

https://github.com/open-rmf/rmf_demos/assets/30265743/14296e3e-0a1c-4353-9536-8cfc62427d24

mxgrey:
@luca-della-vedova Good catch, I've found a subtle bug in the schedule update code that wasn't correctly accounting for planned waits. If you run the latest version the problem should be completely gone now.

@xiyuoh I've added traffic checkpoints for the activity (in this case cleaning) traffic schedule, so now the traffic negotiations should work better.

rmf_demos_fleet_adapter/README.md
```diff
@@ -162,9 +162,9 @@ Request URL: `http://127.0.0.1:22011/open-rmf/rmf_demos_fm/stop/?robot_name=tiny
 ```
 
 ### 4. Send Task Request
-The `start_task` endpoint allows the fleet adapter to send task requests to a specified robot. This endpoint requires a Request Body and a `robot_name` query parameter.
+The `start_activity` endpoint allows the fleet adapter to send task requests to a specified robot. This endpoint requires a Request Body and a `robot_name` query parameter.
 
-Request URL: `http://127.0.0.1:22011/open-rmf/rmf_demos_fm/start_task/?robot_name=tinyRobot1`
+Request URL: `http://127.0.0.1:22011/open-rmf/rmf_demos_fm/start_activity/?robot_name=tinyRobot1`
```
Yadunund, line 167: Nit: Should we update the response body to indicate the existence of `msg["data"]["path"]`?

Yadunund (approved):
It's really awesome seeing how simple it is now to write a new fleet adapter!

rmf_demos_fleet_adapter/README.md
mxgrey: Great idea, added in [39ba81a](https://github.com/open-rmf/rmf_demos/pull/188/commits/39ba81a198ca3effe3e7aaef24b57a5e6c5745d3)

PR #200, opened 2023-12-20
Support one_of for go_to_place

mxgrey:


Yadunund:
Hi @mxgrey is this meant to superseed https://github.com/open-rmf/rmf_demos/pull/198?

mxgrey:
Thanks for pointing that out, I hadn't noticed it. But I do think the changes in this PR make more sense since it involves less code duplication. We already have a distinction between `dispatch_patrol` and `dispatch_go_to_place` so I think lumping the "choose nearest" feature into the latter makes sense.

Yadunund:
Sounds good. Thanks for clarifying.

Yadunund (dismissed):

Yadunund (approved):

PR #206, opened 2024-01-23
Port to harmonic

luca-della-vedova:
### Implemented feature

Depends on https://github.com/open-rmf/rmf_traffic_editor/pull/483.
This PR ports the codebase to Harmonic and drops support for Gazebo classic.
Currently blocked on 24.04 being released and available in the buildfarm, as well as Harmonic rosdep keys being released.

Several parts were cleaned up:

* `rmf_demos_maps` now only needs to build one map, which should speed up its build by a fair bit.
* `gz_classic` plugins and references have been removed.
* `ignition` renamed to `gazebo`.
* `TPE` support has been dropped.
* Default version for gazebo bumped from 6 (fortress) to 8 (harmonic).


mxgrey (approved):
Tested this against 24.04 and rolling. I just needed to tweak the package manifest to correct some dependencies. Now this seems to work great.

Now we just need to do some minor fixes to https://github.com/open-rmf/rmf_simulation/pull/114 and ideally sort out the CI situation, then we can simultaneously merge all the related PRs.

mxgrey:
There seems to be a [CI issue with ignition-cmake2](https://github.com/open-rmf/rmf_demos/actions/runs/9284673912/job/25547540912?pr=206#step:8:27613). My immediate guess is that we're not using the correct dependency key in one or more of the `package.xml` files for `rmf_simulation`.

luca-della-vedova:
> There seems to be a [CI issue with ignition-cmake2](https://github.com/open-rmf/rmf_demos/actions/runs/9284673912/job/25547540912?pr=206#step:8:27613). My immediate guess is that we're not using the correct dependency key in one or more of the `package.xml` files for `rmf_simulation`.

`rmf_simulation` won't really work on rolling / noble until its upstream PR is merged, so this is a bit of a chicken and egg problem.
I'll merge it now and trigger a rerun here, that should fix the CI.

mxgrey (approved):

PR #212, opened 2024-02-06
Add proto-reservation node as core part of RMF

arjo129:
### Implemented feature

Adds a proto-reservation system

Depends on:
* open-rmf/rmf_ros2#325
* open-rmf/rmf_internal_msgs#64

Just adds a line to start the "chope_node" (choping is the Singaporean act of reserving a table usually using some item)


mxgrey (approved):
There was a weird amount of linting that needed to be addressed on old code. I guess noble introduced some changes to the linting tools.

There are some very useful scripts added with this PR, so thanks for that.

We may want to consider just enabling the reservation system for all of the demo worlds, but the office world is a good one to start out with.

Hapy-Capy:
This is very very minor, but this PR reverted the changes made to 'office.building.yaml'  in #263.

PR #225, opened 2024-04-22
Add attach cart action

xiyuoh:
This PR adds attach/detach cart actions to the demo fleet adapter.


rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -183,12 +186,29 @@ def message(data):
...
+        self.mode_pub = self.create_publisher(
+            ModeRequest,
+            'robot_mode_requests',
+            qos_profile=publisher_qos)
+
+        self.action_completed_pub = self.create_publisher(
+            ModeRequest,
+            'action_execution_notice',
+            qos_profile=qos_profile_system_default,
+        )
```
luca-della-vedova, line 210: Can't say I'm too familiar with `action_execution_notice`, but is it needed? What does it do in this case?

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -342,10 +367,24 @@ async def start_activity(
...
+                    task_level_name = \
+                        self.process_waypoints[task.task]['level_name']
+                    for wp in task_wps:
+                        target_loc = Location()
+                        target_loc.x = wp[0]
+                        target_loc.y = wp[1]
+                        target_loc.yaw = wp[2]
+                        target_loc.level_name = task_level_name
+                        path_request.path.append(target_loc)
+                        previous_wp = wp
```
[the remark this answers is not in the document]
xiyuoh: Removed in [5a9e9d9](https://github.com/open-rmf/rmf_demos/pull/225/commits/5a9e9d9bb4e4b1b6629faba494c83eb164fc72e7)

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -100,6 +102,7 @@ def __init__(self, config, nav_path):
         self.config = config
         self.fleet_name = self.config['rmf_fleet']['name']
         mgr_config = self.config['fleet_manager']
+        self.ignore_speed_limit = mgr_config.get('ignore_speed_limit', False)
```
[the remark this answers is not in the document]
xiyuoh: This is relevant if we want to use specific building map nav graphs for both deployment and sim, but do not want to modify any speed limit configured. This flags allow the simulation to ignore any speed limits applied for reasons specific to the deployment.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -226,10 +227,14 @@ def navigate(self, destination, execution):
...
-            )
-            return
+            dock_json = json.loads(destination.dock)
+            if dock_json.get('dock_name') == 'enter_lift':
+                enter_lift = True
+            if not enter_lift:
+                self.attempt_cmd_until_success(
+                    cmd=self.perform_docking, args=(destination,)
+                )
+                return
```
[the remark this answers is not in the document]
xiyuoh: Removed in [5a9e9d9](https://github.com/open-rmf/rmf_demos/pull/225/commits/5a9e9d9bb4e4b1b6629faba494c83eb164fc72e7)

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
xiyuoh: `action_execution_notice` is published by the fleet manager to let the fleet adapter know that the robot has completed the perform action that it is carrying out, so that the fleet adapter can mark the action as finished.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -368,10 +407,48 @@ async def toggle_teleop(robot_name: str, mode: Request):
...
+        async def toggle_attach(robot_name: str, cmd_id: int, mode: Request):
+            response = {'success': False, 'msg': ''}
+            if robot_name not in self.robots:
+                return response
+            # Toggle action mode
+            if mode.toggle:
+                self.attach_cart(robot_name, cmd_id)
+            else:
+                self.detach_cart(robot_name, cmd_id)
+            # self.robots[robot_name].perform_action_mode = mode.toggle
```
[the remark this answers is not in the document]
xiyuoh: Removed in [7c4b0b5](https://github.com/open-rmf/rmf_demos/pull/225/commits/7c4b0b510783fe71aedefcc24389a619a921d52f)

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -342,10 +366,10 @@ async def start_activity(
...
-            path_request.fleet_name = self.fleet_name
-            path_request.robot_name = robot_name
-            path_request.task_id = str(cmd_id)
-            self.path_pub.publish(path_request)
+                path_request.fleet_name = self.fleet_name
+                path_request.robot_name = robot_name
+                path_request.task_id = str(cmd_id)
+                self.path_pub.publish(path_request)
 
             if self.debug:
```
luca-della-vedova, line 374: This doesn't quite seem right, since spaces are significant in Python this would do a bunch of unnecessary assignments and especially publish a message for point in the path array

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -475,6 +553,15 @@ def get_robot_state(self, robot: State, robot_name):
...
             data['replan'] = False
+        if (robot.state.mode.mode == RobotMode.MODE_ACTION_COMPLETED):
+            self.get_logger().info(f'Robot [{robot_name} completed attaching/detaching cart!')
+            completed_cmd_id = 0
+            msg = self._make_mode_request(robot_name, completed_cmd_id,
+                                          RobotMode.MODE_IDLE)
+            # Mark action execution as finished
+            self.action_completed_pub.publish(msg)
+            # # Request for robot idle
+            self.mode_pub.publish(msg)
```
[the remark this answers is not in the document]
xiyuoh: The robot mode might be in `IDLE` mode before the attachment has started/is complete on the simulation end. Iirc the first iteration used `IDLE` as an indication that the attachment is complete, but due to the frequency of the robot state update the fleet manager marked the action as completed almost immediately, so we now have a separate mode to indicate action completion.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
xiyuoh: Thanks for catching that, reverted in [ce846a1](https://github.com/open-rmf/rmf_demos/pull/225/commits/ce846a1ca15fdac7b01667d6564c1bdd42225852)

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
```diff
@@ -475,6 +533,16 @@ def get_robot_state(self, robot: State, robot_name):
...
         else:
             data['replan'] = False
+        if (robot.state.mode.mode == RobotMode.MODE_ACTION_COMPLETED):
+            self.get_logger().info(
+                f'Robot [{robot_name} completed performing its action')
+            completed_cmd_id = 0
+            msg = self._make_mode_request(robot_name, completed_cmd_id,
+                                          RobotMode.MODE_IDLE)
+            # Mark action execution as finished
+            self.action_completed_pub.publish(msg)
```
luca-della-vedova, line 543: While digging at the code I noticed we would publish an message to denote action completion here for every type of action. This is fine but I noticed that in the subscription in the fleet adapter side we still have `teleop` specific code, and when a message is received here we trigger a toggle_teleop call https://github.com/open-rmf/rmf_demos/pull/225/files#diff-1fe77ef1a1d2d883027183fc27ba21f5462e2cb461be28f986296a28eebf6085R283-R285, it seems like this is not the correct behavior and we should only toggle teleop when the action that was being performed is indeed teleop?

rmf_demos_tasks/rmf_demos_tasks/dispatch_cart_delivery.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self, argv=sys.argv):
+        super().__init__('task_requester')
+        parser = argparse.ArgumentParser()
+        parser.add_argument('-g', '--go_to', required=True,
+                            type=str,
+                            help="Go to start waypoint")
+        parser.add_argument('-p', '--pickup', required=True,
+                            type=str,
+                            help="Pickup waypoint name")
```
luca-della-vedova, line 47: The task now requires both `-g` for the place to go to and `-p` for the waypoint pickup name, however the `pickup` argument is not used.
I guess we should either add it as a parameter to the perform action (that however would do nothing in the current sim implementation) or remove one of them altogether

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
xiyuoh: Yup you're right, added a check for an existing `Teleoperation` object stored before calling the `toggle_teleop` endpoint in [04a6fb9](https://github.com/open-rmf/rmf_demos/pull/225/commits/04a6fb939e3e206e67d56f5d31e444f3d96ed7b1). A robot can only perform one action at a time, so if `self.teleoperation` is not None, it means that the current ongoing action is a teleop action. Else, it'd be some other perform action.

luca-della-vedova:
The only thing I would like (as a followup PR probably to avoid this getting stalled further) is a demo scenario showing this feature, maybe having a model for a trolley added to `rmf_demos_assets` or the Gazebo app portal being spawned in a scenario and an example in the README on how to send a task to deliver a cart using perform_action

mxgrey:
> I'd still like another set of eyes, at least for the high level implementation

At the high level the implementation lines up with what I would expect.

It does prompt me to think about how we could improve the way we talk to the simulated robots (the need to add a new field to RobotMode raises some questions about message stability), but that concern is beyond the scope of this PR.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py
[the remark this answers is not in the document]
mxgrey: There's definitely a strong benefit to making `ACTION_COMPLETED` an explicit mode.

In theory if the slotcar plugin could ensure that it **always** switches its mode to `PERFORMING_ACTION` within the same state update that it changes its `task_id` to match the command, then the combination of the correct `task_id` + `IDLE` mode would be enough to infer that it finished the action. But this perpetuates the anti-pattern we've already had within `rmf_fleet_msgs` of relying on implicit combinations of field values to have special meaning. That anti-pattern has always been error prone, and it was the biggest motivation for moving away from `rmf_fleet_msgs` as the API of RMF ~5 years ago.

At this point these messages should only be getting used for slotcar, so I think we should just tweak them as necessary until we have an opportunity to migrate to a more sensible interface for the simulated robots.

PR #235, opened 2024-06-04
Remove rmf_demos_panel

### LINKED ISSUE #234: Drop rmf_demos_panel from main
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

Long overdue. The `rmf_demos_panel` was meant to be an interim dashboard until [rmf-web](https://github.com/open-rmf/rmf-web) was stable which it has been for a couple years now.

The package is unmaintained and we still get a lot of issues related to it. We should delete it for jazzy.

luca-della-vedova:
Fixes #234 

Remove `rmf_demos_panel` and all its mentions. I can't quite place `rmf_demos_dashboard_resources`, it seems it still can be used to generate configurations for `rmf-web` but it hasn't seen updates in some time. For example, its [CMakeLists](https://github.com/open-rmf/rmf_demos/blob/5ab6a3ddd8681da188610a5784bacf1b0eb3cf5f/rmf_demos_dashboard_resources/CMakeLists.txt#L30-L36) doesn't include some of the newer maps (i.e. campus)


README.md
```diff
@@ -54,29 +54,6 @@ By specifying `server_uri`, the fleetadapter will update `rmf-web` `api-server`
 * [Campus World](#Campus-World)
 * [Manufacturing & Logistics World](#Manufacturing-&-Logistics-World)
 
-**RMF Panel**
```
Yadunund, line 57: Can instead direct users to `rmf-web` here?

Yadunund:
@aaronchongth could you comment on whether `rmf_demos_dashboard_resources ` resources is still used in rmf-web?

aaronchongth:
yup `rmf_demos_dashboard_resources` can be used in `rmf-web`. If the intention is to release all packages from `rmf_demos`, I'll redirect users to use the one in the `humble` branch, so this can be removed in `main`.

Meanwhile, it might be a better idea to have it's own repo, or be an orphan branch in `rmf_demos` too

README.md
luca-della-vedova: `rmf-web` is already introduced just before this paragraph, so that would be a bit redundant

luca-della-vedova:
Done! https://github.com/open-rmf/rmf_demos/pull/235/commits/ab3ade14078aa90d4c4b49ff002bbf5f2b6877ff

Yadunund (approved):

PR #230, opened 2024-05-14
Adding server_uri for dispatcher node

aaronchongth:
### Implemented feature

Related to https://github.com/open-rmf/rmf_ros2/pull/355.
Gives the dispatcher node websocket access to servers

PR #237, opened 2024-06-05
Use same empty string default value as fleet adapter launches

aaronchongth:
### Fixed bug

Fix issue mentioned in https://github.com/open-rmf/rmf_demos/pull/230, which causes undesired logging when server URI is not found by using same empty string default vaue as fleet adapter launches


mxgrey (approved):
Thanks for responding to the issue so quickly!

Issue #250, opened 2024-07-16
[Bug]: Rmf demos task for cleaner robot in rmf_demos_fleet adapter mentioned task not found in code and resulting in stopping fleet adapter


techmanjeet:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

2.4.0

### ROS distribution

Rolling

### ROS installation type

Source build

### Other ROS installation methods

jazzy

### Description of the bug

https://github.com/open-rmf/rmf_demos/blob/d6ae16da51940a3406b5c89eab2b50a21a753340/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_manager.py#L350

task is not defined resulting in failure of testing demo task for clean action.

### Steps to reproduce the bug

ros2 launch rmf_demos_gz hotel.launch.xml

### Expected behavior

the clean operation should work

### Actual behavior

its failing due to a wrong variable, that was my finding

Yadunund:
@techmanjeet 
It's not clear to me what issue you're encountering. Your steps to reproduce is incomplete.

> Expected behavior
the clean operation should work
> Actual behavior
its failing due to a wrong variable, that was my finding

What command are you running to issue a task?  Can you share terminal output with the error in a [gist](https://gist.github.com/)? 

Also your setup is a bit confusing. You mentioned you're building both ROS and Open-RMF from source on Ubuntu 24.04 but it's unclear which ROS distro you're using: you highlighted both Rolling and Jazzy.

arjo129:
I was able to duplicate this bug. It is a real bug. Essentially to reproduce, start hotel world. Then run:
```
ros2 run rmf_demos_tasks dispatch_clean -cs clean_lobby -F cleanerBotA -R cleanerBotA_1 --use_sim_time
```
The task is assigned correctly. 

The fleet adapter will not start the activity:
![image](https://github.com/user-attachments/assets/df0c418a-d1d3-434f-ac3a-89f5f9fe4d8c)

PR #251, opened 2024-07-17
Addresses #250: Fixes undefined variable in `FleetManager`

arjo129:
Thanks  @techmanjeet for spotting this issue. I can confirm that this
issue is an issue. Seems like we dont define `task` anymore.


Yadunund (approved):
LGTM.

Issue #253, opened 2024-07-26
[Bug]: RobotCommandHandle crashes when mode_request_cb() is called with missing robot_name/fleet_name


DoppiaEffe94:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 22.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

70b79e25a094ae00e4a75a04593d063823cf3c15

### ROS distribution

Humble

### ROS installation type

Binaries

### Description of the bug

In the RobotCommandHandle.py, `mode_request_cb()` is the callback of the subscriber to the topic `/action_execution_notice`.
This cb function is supposed to check if the fleet_name or robot_name have a value.
However, in python3, this verification is done with the `not ...` operation.

### Steps to reproduce the bug

1. Send a message to the topic `/action_execution_notice` without robot_name or fleet_name.

### Expected behavior

The mode_request_cb() function should return without crashing.

### Actual behavior

The mode_request_cb() function crashes whenever the step one of the described field is called.

xiyuoh:
Hello @DoppiaEffe94 ! Thanks for bringing this up and opening the PR for it. I've tried your steps to reproduce the bug in Office world (22.04 + Humble binaries + RMF source build) but I can't seem to replicate it.

If an empty `fleet_name` is provided, the message would be ignored as `""` is not the fleet name (second condition in [this line](https://github.com/open-rmf/rmf_demos/blob/humble/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py#L724)). If an empty `robot_name` is provided, the message would be accepted and the action is marked as finished successfully.

However, I think it is very reasonable to check for a `not ...`  over a `is None` to rule out empty strings.

Additionally, as I was testing this I've found a relevant bug, which is the lack of a `robot_name` filter inside this `mode_request_cb`. If the action is performed by `tinyRobot2` and I send a message to `/action_execution_notice` and populate it with `robot_name: tinyRobot1`, it would end `tinyRobot2`'s action as well.

To follow up on this issue, may I request for 2 things:
1. If your `mode_request_cb()` is still crashing, could you provide some logs so that I can try to reproduce the problem? Preferably with a link to a gist. This is so that we can check if there are other underlying issues causing the crash.
[...]

PR #254, opened 2024-07-26
Fixed mode_request_cb() when missing robot_name/fleet_name and changed print() into node.get_logger().

DoppiaEffe94:
### Fixed bug

#253 

### Fix applied

In the RobotCommandHandle.py, `mode_request_cb()` is the callback of the subscriber to the topic `/action_execution_notice`.

This cb function is supposed to check if the fleet_name or robot_name have a value.
However, in python 2 and python3, this verification is done with the `not ...` operation.

This PR fixes the if() condition in `mode_request_cb()`, as accordingly to the python2 and python3, considering the fields as Boolean variables.

### DIFF
--- a/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py
+++ b/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/RobotCommandHandle.py
@@ -722,7 +722,7 @@ class RobotCommandHandle(adpt.RobotCommandHandle):
 
     def mode_request_cb(self, msg):
         if not msg.fleet_name or msg.fleet_name != self.fleet_name or\
-                not msg.robot_name:
+                not msg.robot_name or msg.robot_name != self.name:
             return
         if msg.mode.mode == RobotState.IDLE:
             self.complete_robot_action()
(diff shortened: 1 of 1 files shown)


xiyuoh:
Hey @DoppiaEffe94 , thank you for the PR!

As you can see the workflow for DCO has failed, this is because for RMF repositories we need the commits to be signed. This applies to your current and future commits. Could you follow the instructions [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/checking-for-existing-gpg-keys) and [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to create a signing key and sync it to your Github account, then rebase your branch with the signoff? You may want to click on `Details` next to the failed DCO run for more info.

DoppiaEffe94:
Hello @xiyuoh !

I apologize for the DCO problem, I didn't know it was missing.

I added the line `msg.robot_name != self.name`, as suggested, and restored the `self.node.get_logger().info()` into a normal `print()`.

Yadunund:
Just noticed that this was merged into `humble` branch and not `main`. 

Our general policy is to merged fixes first into `main` and backport them to other distro branches.

I don't think we need to revert this PR, but could you open another PR to target `main`?

xiyuoh:
@Yadunund this PR doesn't affect `main` as the fix is in RobotCommandHandle.py while we're using EFC adapter on `main`. The fix is also not applicable to `main` as there is already a [proper check](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py#L473-L475) to make sure that the message is ignored if robot name is empty or `None` that was missing in the `humble` adapter.

Yadunund:
Ah that's great to hear!

PR #255, opened 2024-07-29
Provide example for robot-specific finishing request

xiyuoh:
This PR depends on https://github.com/open-rmf/rmf_ros2/pull/379 and demonstrates how we can specify parking spots for robots in a fleet with Office world.


luca-della-vedova:
I'm a bit on the fence on this, on one side it's definitely useful to showcase the feature, on the other is there a risk the robot might run out of battery if it defaults to parking on a spot that is not a charger?

luca-della-vedova:
But I guess that's a tradeoff with showing the capability :sweat_smile:

PR #256, opened 2024-07-30
update to new dashboard config

koonpeng:
Update the dashboard configs to the new format. Some differences with the new config

* Config must be provided in full, there are many configs in the old format that will fallback to default values, these are no longer valid, all configs are required and there is no fallback.
* Resources are no longer bundled. In the old format, resources must be added to the dashboard source code and bundled together, with the new format, resources can now be external, allowing use cases where it is served from a cdn etc. Resources can still be bundled by putting them in the `public` directory, but there is no longer an import script.
  * The config in the demo now links directly to github so there is no longer any need to "import" them, you can just copy `app-config.json` over.
* Removed dispensers and ingestors resources. They are not implemented in the new map and the old format has always been a hack. In the future, if we want to show them on the map, their coordinates should be included in the building map.


aaronchongth (commented):
LGTM! This should probably be updated only when the new main branch of rmf-web is released

PR #257, opened 2024-08-06
Update dashboard config jsons according to rmf-web#988

aaronchongth:
### Implemented feature

Update dashboard config jsons according to https://github.com/open-rmf/rmf-web/pull/988

### DIFF
--- a/rmf_demos_dashboard_resources/airport_terminal/app-config.json
+++ b/rmf_demos_dashboard_resources/airport_terminal/app-config.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "./app-config.schema.json",
+  "$schema": "https://raw.githubusercontent.com/open-rmf/rmf-web/main/packages/dashboard/app-config.schema.json",
   "rmfServerUrl": "http://localhost:8000",
   "trajectoryServerUrl": "ws://localhost:8006",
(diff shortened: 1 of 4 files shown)


rmf_demos_dashboard_resources/airport_terminal/app-config.json
```diff
@@ -1,9 +1,8 @@
 {
+  "$schema": "./app-config.schema.json",
```
koonpeng, line 2:
```suggestion
  "$schema": "https://raw.githubusercontent.com/open-rmf/rmf-web/main/packages/dashboard/app-config.schema.json"
```

koonpeng (changes requested):

rmf_demos_dashboard_resources/airport_terminal/app-config.json
aaronchongth: hmm, I guess we might still change the location of where the config should be copied too, yeah an absolute url makes more sense, thanks for the suggestion, https://github.com/open-rmf/rmf_demos/pull/257/commits/b3ef3b002b58077a92f201ee4783d465fa0f79da

koonpeng (approved):

PR #262, opened 2024-09-25
Use environment hooks for plugins / assets

luca-della-vedova:
### Implemented feature

Use environment hooks instead of manually creating environment variables for launching gazebo simulations.

Needs https://github.com/open-rmf/rmf_simulation/pull/136.
Brings us 90% of the way there to only using hooks for setting variables (the remaining TODO is not fully straightforward and might require changes on `rmf_traffic_editor` so left it for now, more explanation below).
Both plugins (with upstream `rmf_simulation` PR) and rmf_demos_assets (with this PR) will set environment variables to make lookups easier.
This reduces boilerplate on the launch file and also makes it easier for third parties to use these packages without having to manually set their environment variables.

There is a remaining TODO because for each map we add to the path:

```
(find-pkg-share $(var map_package))/maps/$(var map_name)/models
```

This makes it possible for the root level world file to be able to lookup the level meshes that come in the format (example from office world):

```
<include>
  <name>building_L1</name>
  <uri>model://building_L1</uri>
  <pose>0 0 0.0 0 0 0</pose>
</include>
```

However, if we want this to work with hooks and, for example, add `find-pkg-share $(var map_package))/maps` to an environment hook, the `uri` should change to the subfolder which doesn't quite look as clean. I'm also not 100% sure it would be able to find its submeshes:

```
<uri>model://office/models/building_L1</uri>
```

For this reason I left this as a TODO, I also left the `.gazebo/models` here since it's probably not a good idea to add that to the path of users when our packages are built

### DIFF
--- a/rmf_demos_assets/CMakeLists.txt
+++ b/rmf_demos_assets/CMakeLists.txt
@@ -4,6 +4,8 @@ project(rmf_demos_assets)
 
 find_package(ament_cmake REQUIRED)
 
+ament_environment_hooks("${CMAKE_CURRENT_SOURCE_DIR}/hooks/${PROJECT_NAME}.dsv.in")
+
 install(
   DIRECTORY models
   DESTINATION share/${PROJECT_NAME}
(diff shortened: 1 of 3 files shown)


arjo129 (approved):
Tested. Seems to be working correctly.

PR #263, opened 2024-10-04
Remove nav graph for fleet manager and update Office world

xiyuoh:
The current Fleet Manager node takes in two arguments: fleet config file and navigation graph file. The latter is not being used anywhere throughout the code. This PR removes this argument from the script and the fleet adapter launch file. This would also allow us to use the fleet manager node for fleets without navigation graphs, e.g. read-only fleets.

Additionally this PR also slightly modifies the current Office world. Currently there exists a path between waypoints `patrol_A1` and `patrol_B`, but in Gazebo the path is blocked by chairs and MiR models. [73cd00f](https://github.com/open-rmf/rmf_demos/pull/263/commits/73cd00fe0868b997f6082e69ae5ce14e594626aa) modifies the nav graph lanes such that the robots can go to `patrol_B` using an unblocked path. Some of the Drawer models along the bottom of the Office world were also removed/moved to make space for robot patrol.

Before:
office_world_before

After:
office_world_after

### DIFF
--- a/rmf_demos_fleet_adapter/launch/fleet_adapter.launch.xml
+++ b/rmf_demos_fleet_adapter/launch/fleet_adapter.launch.xml
@@ -11,7 +11,7 @@
   <!-- Fleet manager -->
   <node pkg="rmf_demos_fleet_adapter"
         exec="fleet_manager"
-        args="--config_file $(var config_file) --nav_graph $(var nav_graph_file)"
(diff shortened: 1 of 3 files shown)


luca-della-vedova (approved):
Ah yes the cluttered office. I _believe_ this was done for a demonstration where nav2 was trained in simulation and the map used in real life, luckily the office became less cluttered since then!

PR #264, opened 2024-10-08
Adds the ability to set the update rate of the simulation

arjo129:
### Implemented feature

This commit adds the ability to set the update rate of simulation to be faster than real time via the `update_rate` option.

For instance if you want a 5x real time factor simulation running via CLI simply add the flags:

```
ros2 launch rmf_demos_gz office.launch.xml update_rate:=500
```
Note by default this sets gazebo's internal step size. Usually we have 100 steps per second of simulation, hence if we want 5x RTF we need to set update_rate:=500. Also note there is a bug in gazebo which also results in the ecm inspector showing the wrong "RTF" if I use the `-z` flag.

### DIFF
--- a/rmf_demos_gz/launch/airport_terminal.launch.xml
+++ b/rmf_demos_gz/launch/airport_terminal.launch.xml
@@ -4,6 +4,7 @@
   <arg name="use_sim_time" default="true"/>
   <arg name="failover_mode" default="false"/>
   <arg name="gazebo_version" default='8'/>
+  <arg name="update_rate" default='1000'/>
 
   <!-- Common launch -->
   <include file="$(find-pkg-share rmf_demos)/airport_terminal.launch.xml">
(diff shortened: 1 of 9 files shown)


Yadunund (changes requested):
imo  `update_rate` sounds very ambiguous since it's not clear what this is referring to along with the units. There is room to think it updates the rate of the fleet adapter updates for example. How about `sim_update_rate`?

luca-della-vedova:
The default step size for rmf simulations is [0.01](https://github.com/open-rmf/rmf_traffic_editor/blob/71e419c9cda7fa17d47055ded3445859c542979e/rmf_building_map_tools/building_map/templates/gz_world.sdf#L5) though, not 0.001, does the command line argument also set the step size to 0.001 behind the scenes or would this make the simulation run at 10x RTF by default?

arjo129:
> The default step size for rmf simulations is [0.01](https://github.com/open-rmf/rmf_traffic_editor/blob/71e419c9cda7fa17d47055ded3445859c542979e/rmf_building_map_tools/building_map/templates/gz_world.sdf#L5) though, not 0.001, does the command line argument also set the step size to 0.001 behind the scenes or would this make the simulation run at 10x RTF by default?

You are right this is my mistake, I've changed the defaults.

Yadunund (approved):

PR #272, opened 2024-12-27
Change robot types to include a namespace

luca-della-vedova:
## New feature implementation

Depends on https://github.com/open-rmf/rmf_traffic_editor/pull/521.

### Implemented feature

This PR changes the `robot_spawn_type` in demo maps to include an `Open-RMF` namespace, similar to how normal models have an `OpenRobotics` namespace. The relevant models (all except the `RobotPlaceholder` that is not used anymore as far as I can tell) have been uploaded to a new [Open-RMF organization](https://app.gazebosim.org/Open-RMF) on the gazebosim portal.
This is to help in future-proofing the maps for the site editor.

### Test it!

When doing a clean build with this branch and the matching `rmf_traffic_editor` branch, all demos should still work.


rmf_demos_maps/maps/airport_terminal/airport_terminal.building.yaml
```diff
@@ -727,8 +727,8 @@ levels:
       - {model_name: OpenRobotics/Construction Cone, name: Construction Cone, static: true, x: 306.031, y: 612.471, yaw: 1.5708, z: 0}
       - {model_name: OpenRobotics/Construction Cone, name: Construction Cone, static: true, x: 329.259, y: 609.5069999999999, yaw: 1.5708, z: 0}
       - {model_name: OpenRobotics/MopCart3, name: MopCart3, static: false, x: 283.704, y: 443.583, yaw: 1.5708, z: 0}
-      - {model_name: OpenRobotics/TeleportDispenser, name: mopcart_dispenser, static: true, x: 284.29, y: 443.7, yaw: 1.5708, z: 0}
-      - {model_name: OpenRobotics/TeleportIngestor, name: mopcart_collector, static: true, x: 317.372, y: 589.088, yaw: 1.5708, z: 0}
+      - {model_name: Open-RMF/TeleportDispenser, name: mopcart_dispenser, static: true, x: 284.29, y: 443.7, yaw: 1.5708, z: 0}
+      - {model_name: Open-RMF/TeleportIngestor, name: mopcart_collector, static: true, x: 317.372, y: 589.088, yaw: 1.5708, z: 0}
```
luca-della-vedova, line 731: Note that this change, together with uploading the `TeleportIngestor` and `TeleportDispenser` plugins to fuel, will make `pit_crew` download the models and there will be silently two of the same model in `.gazebo/models` as well as the `rmf_demos` installation folder.
The possible actions are either:
1) Remove the models from Fuel (Site editor won't be able to download it) or:
2) Remove the model from `rmf_demos_assets` and it will be downloaded by pit_crew (it's just a [box with a plugin](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_assets/models/TeleportIngestor/model.sdf)).
3) Keep as is and there might be ambigious lookup in Gazebo.

Option 3) is definitely not a good idea, 1) and 2) have different tradeoffs

mxgrey: Is gazebo only given the plain model name without the organization name to distinguish between them? And is gazebo not able to account for an organization name when searching for a model?

If yes to both then I imagine this is a broader problem that will need to be resolved within Gazebo or else it's going to cause trouble for large projects in general.

luca-della-vedova: > Is gazebo only given the plain model name without the organization name to distinguish between them? 

Yes. As part of the generation the organization name is removed [here](https://github.com/open-rmf/rmf_traffic_editor/blob/cb95b26000df592e50b092c9b54191361d86aec7/rmf_building_map_tools/building_map/generator.py#L31-L36) so `OpenRobotics/Chair` is translated into a bare `model://Chair`.

> And is gazebo not able to account for an organization name when searching for a model?

It should definitely be possible if we specify models the "proper gazebo way", which is through Fuel URLs other than our own schema, like [this](https://github.com/gazebosim/gz-sim/blob/9d6230a562c5d463306070b07269b135f9ecd134/examples/worlds/fuel.sdf#L62-L66).


I haven't looked into what happens if we remove the "organization stripping" logic from our world generation, but there are a few caveats:

* If it works as intended, we will need to change our model download pipeline to have subfolders for organizations (it currently doesn't).
* We will also need to change how we install models from `rmf_demos_assets` to be installed under an `Open-RMF` organization subfolder.
* If users have their own asset package (i.e. forked `rmf_demos_assets`) it will stop working unless they also update their folder structure.

Finally, the root of this specific issue is that the asset exists both on Fuel and `rmf_demos_assets` and that would persist. `pit_crew` has no knowledge of what models have been installed and sourced in the workspace so it can't detect that an `Open-RMF/TeleportIngestor` has been installed. It will only detect that the map has an `Open-RMF/TeleportIngestor` model and thus it should try downloading it.

luca-della-vedova: I removed the `TeleportIngestor` and `TeleportDispenser` models from Fuel to avoid this issue for now.
This means site editor will still not be able to fetch them but at least there won't be risk of duplicated models.

luca-della-vedova: I also reverted moving the `TeleportIngestor` and `TeleportDispenser` plugins to the `Open-RMF` org in https://github.com/open-rmf/rmf_demos/pull/272/commits/ee7cd84b624b66e004e94e7daee427c5fe8f7a2f (and the matching [PR](https://github.com/open-rmf/rmf_traffic_editor/pull/521/commits/93255b2f31cefd22f58e2f5af6ec100053adfb6b) in rmf_traffic_editor)), this should reduce further the potential impact here

mxgrey (approved):

PR #289, opened 2025-04-20
set self.teleoperation to None when finish_action

piliwilliam0306:
### Fixed bug

This PR prevents keep calling `self.teleoperation.update` after robot finishes **teleop** action.


mxgrey (approved):
Good catch, thanks for the contribution!

PR #292, opened 2025-04-30
Use 9 as default gazebo_version

mxgrey:
gz-sim 9 is the version of Gazebo that is vendored on the current rolling branch, so we should target that version by default in order to get rmf_demos released on rolling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ These demos were built and tested on
 
 * [Ubuntu 24.04 LTS](https://releases.ubuntu.com/24.04/)
 
-* [ROS 2 - Jazzy](https://docs.ros.org/en/jazzy/Releases/Release-Jazzy-Jalisco.html)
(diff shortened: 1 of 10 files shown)


mxgrey:
The build test isn't passing due to ongoing issues with gz fuel, so I'll override that once the PR is approved.

xiyuoh (approved):
Should we update and/or add a version table in the [System Requirements](https://github.com/open-rmf/rmf_demos#system-requirements) section for users to understand compatibility between RMF demos and Gazebo versions? It could also be helpful to note that they'll can run `main` with Jazzy by adding `gazebo_version:=8`.

mxgrey:
Since Kilted isn't out yet, I've replaced `Jazzy` with `Rolling`, but I've [opened a ticket](https://github.com/open-rmf/rmf_demos/issues/293) to remind us to update that to `Kilted`.

xiyuoh (approved):
Perfect, LGTM!

PR #294, opened 2025-05-05
Generate new cmd_id for stop request

piliwilliam0306:
### Fixed bug

When [stop](https://github.com/open-rmf/rmf_demos/blob/78e736065204d863a956034f778dfccc16b60aee/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py#L258) is called, `path_request` is ignored due to repeated task request in https://github.com/open-rmf/rmf_simulation/blob/7d70ff18e123b99487485acae1692a7276e380cd/rmf_robot_sim_common/src/slotcar_common.cpp#L241
A new `cmd_id` for stop request is generated to fix this issue


luca-della-vedova (approved):
Apologies for the delay in the review! This looks well justified and implemented, I'll give it till end of week if anyone else wants to pitch in with a review but LGTM

mxgrey (approved):

PR #303, opened 2025-05-19
Demos main no longer compatible with rmf_simulation jazzy

xiyuoh:
Updating readme to reflect distro/library compatibility.

Due to https://github.com/open-rmf/rmf_demos/pull/262 (on rmf_demos `main`) depending on https://github.com/open-rmf/rmf_simulation/pull/136 (available in rmf_simulation `main` but not `jazzy`), environment hooks for plugin paths are absent in rmf_simulation `jazzy`, so rmf_demos `main` is no longer compatible with all RMF distros, and has to match rmf_simulation between `main` and `jazzy`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,7 @@ These demos were built and tested on
 * [ROS 2 - Rolling](https://docs.ros.org/en/jazzy/Releases/Release-Rolling-Ridley.html)
 
 * [Gazebo Ionic](https://gazebosim.org/docs/ionic)
-> Note: The `main` branches of the core RMF libraries are fully supported on ROS 2 Humble, Iron, and Jazzy as well, but you will need to use the distro-specific branches for `rmf_traffic_editor` and `rmf_simulation`.
->
-> You can use `rmf_demos` with any ROS distro by explicitly setting the `gazebo_version:=#` launch parameter, replacing `#` with the appropriate version of Gazebo for that ROS distro, e.g. `8` for `Jazzy`.
+> Note: The `main` branches of the core RMF libraries are fully supported on ROS 2 Humble, Iron, and Jazzy as well, but you will need to use the distro-specific branches for `rmf_traffic_editor`, `rmf_simulation` and `rmf_demos`.
 
 ## Installation
 Instructions can be found [here](https://github.com/open-rmf/rmf).


arjo129:
Just curious why do environment hooks not work in jazzy? Is this to do with the way we deal with binaries? Or did we not back-port the env-hooks to jazzy? Or do they not work with thte jazzy binaries?

xiyuoh:
@arjo129 this [commit](https://github.com/open-rmf/rmf_simulation/commit/9e0e5b70d67573ae9519c59d1e27d8b380b57f0e) which environment hooks depend on wasn't backported to `jazzy`. The alternative is we cherry pick this commit into rmf_simulation `jazzy`. Tested it out locally and it seems to work fine.

mxgrey:
It may be reasonable to do the cherry picking to make it compatible with jazzy, but that would still leave it incompatible with humble unless we also cherry-pick into humble, right?

I think for now the most common use case for community members will be people wanting to simulate on humble. It would be great if they can use the `main` branch of `rmf_demos` on humble so they can also use the scripts for the new features on the `main` branches of the core libraries.

arjo129:
Yes. I do think cherry-picking is the way to go. I don't know if I'd count this as a "feature" since it feels like more of an internal refactor. Hooks have been supported for as far as I can remember in ROS2.

xiyuoh:
Thanks both for helping to brainstorm! I'm testing the backports locally and the different combinations for `jazzy` (both `jazzy`, both `main`, sim `jazzy` + demos `main`) are working well.

However things are trickier for `humble`. Since we still use [ign commands](https://github.com/open-rmf/rmf_demos/blob/humble/rmf_demos_gz/launch/simulation.launch.xml#L26) in `humble`, we'd have to check against the ignition/gazebo version used to determine which command to launch Gazebo. It looks to me like if we want to support demos `main` to be used with `humble`, we'd have to re-introduce some of these params in the launch file.

mxgrey:
Thanks for investigating, @xiyuoh . I don't want to make this more complex than it needs to be.

Let's move forward with just backporting it to jazzy, and if anyone in the community cares about simulating the latest features on humble then they can sort out the complexity and open a PR for it.

xiyuoh:
Closing in favor of backport PRs

arjo129:
> Thanks both for helping to brainstorm! I'm testing the backports locally and the different combinations for `jazzy` (both `jazzy`, both `main`, sim `jazzy` + demos `main`) are working well.
> 
> However things are trickier for `humble`. Since we still use [ign commands](https://github.com/open-rmf/rmf_demos/blob/humble/rmf_demos_gz/launch/simulation.launch.xml#L26) in `humble`, we'd have to check against the ignition/gazebo version used to determine which command to launch Gazebo. It looks to me like if we want to support demos `main` to be used with `humble`, we'd have to re-introduce some of these params in the launch file.

As of now the latest fortress version should load `GZ_*` variables: https://github.com/gazebosim/gz-sim/blob/cb6fa87454c0eae212c4db6e53538cb8052f67ac/src/SystemLoader.cc#L49

PR #304, opened 2025-05-20
Use environment hooks for plugins / assets (#262)

xiyuoh:
Backport environment hooks to `jazzy`, as discussed in https://github.com/open-rmf/rmf_demos/pull/303.


mxgrey:
Same thought as on https://github.com/open-rmf/rmf_simulation/pull/144 , we should probably switch to `jazzy` [here](https://github.com/open-rmf/rmf_demos/blob/jazzy/.github/workflows/build.yaml#L21).

luca-della-vedova (approved):

PR #307, opened 2025-06-03
Use builtin gazebo model downloading

luca-della-vedova:
### Implemented feature

Needs https://github.com/open-rmf/rmf_traffic_editor/pull/535, also closes https://github.com/open-rmf/rmf_traffic_editor/issues/534

The upstream PR has most of the description, by using Gazebo itself to download models we won't need custom paths anymore (such as the user's home folder) cutting down on build times and most importantly unblocking distribution of `rmf_demos_maps` binaries.


mxgrey (approved):
The PR makes sense and it's working well. I look forward to having a functioning release of `rmf_demos`!

PR #332, opened 2025-09-26
Add task_assignment_strategy to fleet config

kjchee:
This PR is related to rmf_task open-rmf/rmf_task#129.

It adds a new parameter called: task_assignment_strategy in the fleet_config.

### DIFF
--- a/rmf_demos/config/office/tinyRobot_config.yaml
+++ b/rmf_demos/config/office/tinyRobot_config.yaml
@@ -34,6 +34,7 @@ rmf_fleet:
   responsive_wait: False # Should responsive wait be on/off for the whole fleet by default? False if not specified.
   use_parking_reservations: True
   reassign_task_interval: 120 # seconds, specify how often a task reassignment should be triggered in the fleet
+  task_assignment_strategy: "ShortestFinishTime" # Task planning strategy eg.["ShortestFinishTime", "IdlePreferred"]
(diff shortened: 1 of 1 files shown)


rmf_demos/config/office/tinyRobot_config.yaml
```diff
@@ -34,6 +34,19 @@ rmf_fleet:
   responsive_wait: False # Should responsive wait be on/off for the whole fleet by default? False if not specified.
   use_parking_reservations: True
   reassign_task_interval: 120 # seconds, specify how often a task reassignment should be triggered in the fleet
+  task_assignment_strategy:
+    profile: "DefaultFastest"   # [DefaultFastest, BatteryAware, Custom]
+    # DefaultFastest: Prioritize task efficiency (fastest finish time)
+    # BatteryAware: Optimize throughput with battery SoC in mind
+    # Custom: Use custom weights below
+    weights:  # polynomial coeffs
```
xiyuoh: Suggest to explain how the coefficients are assigned, since users would not know that [constants are ignored](https://github.com/open-rmf/rmf_task/pull/129#discussion_r2521477154) without digging into the relevant PR/code. For example,

```
weights:  # Polynomial coefficients in the format [A, B, ..., N] such that cost = Ax + Bx^2 + ... + Nx^N
```

PR #338, opened 2025-10-24
Add use_site_map param

xiyuoh:
Add `use_site_map` param to toggle between serving maps with legacy `rmf_building_map_tools` or the latest `rmf_site_map_server`

### DIFF
--- a/rmf_demos/package.xml
+++ b/rmf_demos/package.xml
@@ -17,7 +17,6 @@
   <exec_depend>rmf_task_ros2</exec_depend>
   <exec_depend>rmf_fleet_adapter</exec_depend>
   <exec_depend>rmf_building_map_tools</exec_depend>
-  <exec_depend>rmf_site_map_server</exec_depend>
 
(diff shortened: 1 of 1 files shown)


rmf_demos/package.xml
```diff
@@ -17,6 +17,7 @@
   <exec_depend>rmf_task_ros2</exec_depend>
   <exec_depend>rmf_fleet_adapter</exec_depend>
   <exec_depend>rmf_building_map_tools</exec_depend>
+  <exec_depend>rmf_site_map_server</exec_depend>
```
mxgrey: Conventionally it would make sense to add this dependency, but we don't have a good story yet for distributing cargo-based dependencies, so this is going to create quite a lot of friction for users who don't plan on using `rmf_site_map_server`.

What I'm about to suggest is a bit questionable, but I think we should remove this dependency from here and instead just let it exist in the `package.xml` of `rmf_site_demos`.

PR #344, opened 2026-01-13
Switch to using ros_gz_sim to run Gazebo

luca-della-vedova:
### Fixed bug

Brought up in https://github.com/open-rmf/rmf_visualization/pull/88#issuecomment-3741945109.
It seems argument substitution is behaving a bit oddly with our current way to run gazebo through the `executable` tag.
Specifically, on Rolling the `gz_headless` variable seems to break the launch command, as shown in the comment (it also doesn't work for me, but running in headless mode actually works).

### Fix applied

I tried to change the content of the `gz_headless` variable to contain the whole command, change its position in the command, various combination of content but I can't quite figure out where the issue exactly lies. It's really very odd, sometimes only headless mode works, sometimes only normal mode works, but never both at the same time.
By contrast, using `ros_gz_sim` seems to always work. This seems to be also the officially recommended way to launch Gazebo based on the [official docs](https://gazebosim.org/docs/latest/ros2_launch_gazebo/).
### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

I tested this on both a Jazzy and a Rolling machine.

- [ ] I used a GenAI tool in this PR.
- [x] I did not use GenAI

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,12 @@ Start the backend API server via `docker` with host network access, using the de
 
 ```bash
 docker run \
-  --network host \
-  -it \
+  --network host -it --rm \
   -e ROS_DOMAIN_ID=<ROS_DOMAIN_ID> \
   -e RMW_IMPLEMENTATION=<RMW_IMPLEMENTATION> \
-  ghcr.io/open-rmf/rmf-web/api-server:latest
+  ghcr.io/open-rmf/rmf-web/api-server:jazzy-nightly
+
+# Use the appropriate tag for different ROS 2 distributions
 ```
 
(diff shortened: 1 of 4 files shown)


xiyuoh (approved):
Awesome! Tested this for both Jazzy and Rolling, am able to run demos for hotel and office. Thanks for the quick fix!

PR #345, opened 2026-01-13
Bump changelogs and release 2.8.1

luca-della-vedova:
Will retag and bloom-release to rolling afterwards


mxgrey:
I was hoping to get https://github.com/open-rmf/rmf_demos/pull/332 before the next release. If it's too late then I can just follow up with another patch release later.

luca-della-vedova:
I think this is just another manifestation of https://github.com/ros2/rviz/pull/1635#pullrequestreview-3537568103. It should be fixed once the rolling sync is done in a few days.
We could have gone for a "green CI" way and only fix things when we could have got green CI but that would have forced us to wait for the next sync to have the rviz fix in, but `rmf` is failing already and would have been removed from Rolling in the meanwhile.
Instead, we tested in the `testing` ROS repos to use the binaries before they make it to the public and we validated that everything works, so this red CI is expected but a fluke.
And to be honest Rolling has been broken for a long time so this failure is really nothing new, it's just that it wasn't caught by CI because we haven't been running CI for some time.
Looking at the [logs of the builds](https://build.ros2.org/job/Rbin_uN64__rmf_visualization_rviz2_plugins__ubuntu_noble_amd64__binary/buildTimeTrend) in the buildfarm, you can see how it's been red for a long time, and just only became green because we merged and released the fix in `rmf_visualization`.

PR #357, opened 2026-07-15
Add robots dynamically during runtime

jixiancgh:
### Implemented feature

Add robots dynamically during runtime.

Refer to this [issue](https://github.com/open-rmf/rmf_ros2/issues/523). Requires this `rmf_internal_msgs` [PR](https://github.com/open-rmf/rmf_internal_msgs/pull/95).

This PR implements runtime robot registration in the `rmf_demos` fleet adapter and fleet manager, adding a robot to a running fleet without restarting the adapter. This demonstrates how a user can potentially write a similar service callback in their fleet adapter to achieve the same results.

### How to test

I've added a third robot `tinyRobot3` using an existing point in the map as charger/spawn point so the office demo can be used to exercise adding a robot at runtime. We can alternatively create a new demo map for this feature if needed.

Launch office demo
`ros2 launch rmf_demos_gz office.launch.xml`

Call service
`ros2 service call /tinyRobot/add_robot rmf_fleet_msgs/srv/AddRobot \
    "{robot_name: 'tinyRobot3', robot_config: '{\"compatible_chargers\": [\"tinyRobot3_charger\"]}'}"`

### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [x] I used a GenAI tool in this PR.
- [ ] I did not use GenAI

Generated-by: Claude Opus 4.8

### DIFF
--- a/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
+++ b/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
@@ -15,7 +15,6 @@
 import argparse
 import asyncio
 import faulthandler
-import json
 import math
 import sys
 import threading
@@ -33,12 +32,12 @@ from rclpy.qos import QoSReliabilityPolicy as Reliability
 import rmf_adapter
 from rmf_adapter import Adapter
 import rmf_adapter.easy_full_control as rmf_easy
+from rmf_demos_msgs.srv import AddRobot
 from rmf_fleet_msgs.msg import ClosedLanes
 from rmf_fleet_msgs.msg import LaneRequest
 from rmf_fleet_msgs.msg import ModeRequest
 from rmf_fleet_msgs.msg import RobotMode
 from rmf_fleet_msgs.msg import SpeedLimitRequest
-from rmf_fleet_msgs.srv import AddRobot
 import yaml
 
 from .RobotClientAPI import RobotAPI
@@ -524,16 +523,6 @@ def ros_connections(
             f'Received request to add robot [{robot_name}] at runtime'
         )
 
-        try:
-            cfg = json.loads(request.robot_config)
-            if not isinstance(cfg, dict):
-                raise ValueError('robot_config must be a JSON object')
-        except (json.JSONDecodeError, ValueError) as err:
-            response.success = False
-            response.message = f'Invalid robot_config JSON: {err}'
-            node.get_logger().error(response.message)
-            return response
-
         if not api.add_robot(robot_name):
             response.success = False
             response.message = (
@@ -543,9 +532,9 @@ def ros_connections(
             return response
 
         fleet_config.add_known_robot_configuration(
-            robot_name, 
+            robot_name,
             rmf_easy.RobotConfiguration(
-                compatible_chargers=cfg.get('compatible_chargers')
+                compatible_chargers=request.robot_config.compatible_chargers
             )
         )
         robot_config = fleet_config.get_known_robot_configuration(robot_name)
--- a/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
+++ b/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
@@ -534,12 +534,6 @@ def ros_connections(
             node.get_logger().error(response.message)
             return response
 
-        robot_config = rmf_easy.RobotConfiguration(
-            compatible_chargers=cfg.get('compatible_chargers')
-        )
-        fleet_config.add_known_robot_configuration(robot_name, robot_config)
-        robot_config = fleet_config.get_known_robot_configuration(robot_name)
-
         if not api.add_robot(robot_name):
             response.success = False
             response.message = (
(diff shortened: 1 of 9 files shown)


rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -497,6 +508,59 @@ def mode_request_cb(msg):
...
+        fleet_config.add_known_robot_configuration(robot_name, robot_config)
+        robot_config = fleet_config.get_known_robot_configuration(robot_name)
+
+        if not api.add_robot(robot_name):
+            response.success = False
+            response.message = (
+                f'Fleet manager rejected registration of {robot_name}'
+            )
+            node.get_logger().error(response.message)
+            return response
```
xiyuoh: The `api.add_robot(robot_name)` check should come before initializing `robot_config` and adding it to fleet config. Otherwise, if the fleet manager rejects registration of the new robot, we would have already added the new robot config to the existing fleet config.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -497,6 +508,59 @@ def mode_request_cb(msg):
...
+        except (json.JSONDecodeError, ValueError) as err:
+            response.success = False
+            response.message = f'Invalid robot_config JSON: {err}'
+            node.get_logger().error(response.message)
+            return response
+
+        robot_config = rmf_easy.RobotConfiguration(
+            compatible_chargers=cfg.get('compatible_chargers')
+        )
+        fleet_config.add_known_robot_configuration(robot_name, robot_config)
```
xiyuoh: Minor suggestion to reduce confusion/redefining robot_config:

```suggestion
        fleet_config.add_known_robot_configuration(
            robot_name,
            rmf_easy.RobotConfiguration(
                compatible_chargers=cfg.get('compatible_chargers')
            )
        )
```

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
```diff
@@ -497,6 +508,59 @@ def mode_request_cb(msg):
...
+            response.message = 'Invalid request'
+            node.get_logger().error(response.message)
+            return response
+
+        node.get_logger().info(
+            f'Received request to add robot [{robot_name}] at runtime'
+        )
+
+        try:
+            cfg = json.loads(request.robot_config)
```
xiyuoh: I don't know how I feel about embedding JSON into a `.srv` file. If we are adding srv/msg files anyway, I might suggest defining a `RobotConfiguration.msg` that mirrors EasyFullControl's [RobotConfiguration](https://github.com/open-rmf/rmf_ros2/blob/6978eae2a8b022f57bac22cf3102dff271900b7b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp#L266-L272), then replace the JSON robot_config with it.

xiyuoh (changes requested):
As mentioned in https://github.com/open-rmf/rmf_internal_msgs/pull/95 we shouldn't be adding non-core RMF msgs to `rmf_internal_msgs`. A possible alternative is to create a separate `rmf_demos_fleet_msgs` package in this repository.

rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py
jixiancgh: For RobotConfiguration, the Python [binding](https://github.com/open-rmf/rmf_ros2/blob/6978eae2a8b022f57bac22cf3102dff271900b7b/rmf_fleet_adapter_python/src/adapter.cpp#L1005-L1007) only exposes `compatible_chargers`. The other properties like `responsive_wait`, `max_merge_waypoint_distance`, `max_merge_lane_distance` and `min_lane_length` aren't modifiable from Python.

There could be an issue when these parameters are changed at the fleet level in config, and users add a robot through this newly added service and expect it to follow the `fleet_config` params. 

EasyFullControl's `add_robot` gets fleet default `nav_params` and only overrides when the robot config option `has_value` [here](https://github.com/open-rmf/rmf_ros2/blob/6978eae2a8b022f57bac22cf3102dff271900b7b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp#L3637-L3651). However the C++ constructor defaults to 1e-3 / 0.3 / 1e-8 [here](https://github.com/open-rmf/rmf_ros2/blob/6978eae2a8b022f57bac22cf3102dff271900b7b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp#L268-L271) rather than `std::nullopt`, silently overriding the fleet config values when there are no values passed in through the Python binding.

I've included all the fields in the message, but we'll probably need to make changes to `rmf_fleet_adapter_python` to avoid overriding `fleet_config` defaults.

